PR #995, opened 2024-03-06
Cftime support

mnurisso:
## PR description:

Support to `cftime` in `timmean` `Reader` method.

- [ ] `exclude_incomplete `compatible
- [ ] `center_time` compatible
- [ ] `check_time` compatible

Notes:

`np.isnat` is not available
`pandas` offsets have to be rewritten

----

 - [ ] Tests with a `cftime` xarray
 - [ ] Documentation in advanced topics
 - [ ] Changelog is updated.
 - [x] environment.yml and pyproject.toml are updated if needed, together with the lumi installation tool.


oloapinivad:
Apparently following the recent upgrades in pandas is possible to change the datetime precision so that we can access with lower time precision also weird calendars

PR #1649, opened 2025-01-30
Handling plev in Global Biases diagnostic

silviacaprioli:
## PR description:

With this PR, standard 2D bias plots (both total and seasonal) will be generated for 3D variables when one or more pressure levels `plev` are specified.

 - [x] Changelog is updated.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,29 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-24.04
+  tools:
+    # to be checked if this makes sense, EsmVal is using miniconda
+    python: "mambaforge-4.10"
+
+# set up the environment
+conda:
+  environment: environment.yml
+
+# install the local package
+python:
+  install:
+    - method: pip
+      path: .
+      extra_requirements:
+        - docs
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+   configuration: docs/sphinx/source/conf.py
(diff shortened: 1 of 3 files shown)


silviacaprioli:
image

added scientific notation in contours (`fmt = "%1.1e" if max(ds.levels) > 1000 or min(ds.levels) < 0.01 else "%1.1f"`)

src/aqua/util/output_saver.py
```diff
@@ -162,6 +164,9 @@ def generate_name(self, diagnostic_product: str = None, var: str = None, model_2
         parts.extend(additional_parts)
         parts.append(suffix)
 
+        self.logger.debug(f'oooo {ordered_keys}')
```
mnurisso, line 167: Is this something to be left in the PR?

src/aqua_diagnostics/global_biases/cli_global_biases.py
```diff
@@ -142,73 +142,87 @@ def main():
         var_attributes = config["biases_plot_params"]['bias_maps'].get(var_name, {})
         vmin, vmax = var_attributes.get('vmin'), var_attributes.get('vmax')
 
+        if 'plev' in data[var_name].dims and plev:
+            plev_list = [plev] if isinstance(plev, (int, float)) else plev
```
mnurisso, line 146: there's a method `to_list()` to do exactly this that you can use.
`from aqua.util import to_list`

mnurisso (approved):
I tried the notebook to check there was no disruption in the plot routines and everything seems ok. 

@silviacaprioli I think now nothing (but the changelog) is missing here right?
@jhardenberg did you test the branch already? Do we want to have a run tonight with the branch to check that no disruption on other diagnostics is done and the new plots are generated?

src/aqua/util/output_saver.py
silviacaprioli: no it's something left from some debug I did, thanks :)

silviacaprioli:
> @silviacaprioli I think now nothing (but the changelog) is missing here right? 

The changelog has been added. I believe Jost tested it, but that was before all the visual plot changes.

mnurisso:
@jhardenberg this is ready. Let me know if a aqua-web run is required or this can go in the main. Feel free to merge directly if no run is needed.

mnurisso:
As discussed offline this can be merged.

PR #1656, opened 2025-01-31
[OP] Develop more elaborate capacity to read FDB and Bridge info from file

### LINKED ISSUE #1617: Integration of dynamical FDB HPC and DataBridge info coming from the Workflow in the
### Feature request

We sketch here what is the actual result of the internal discussion on how to be able to access updated info on which data have been transferred to the databridge and which have to be read from the HPC FDB.

## Yaml file produced by workflow

The first step is to have a yaml file with this structure

```yaml
# template for yaml info file
hpc:
    data_start_date: 19900101T0000 #experiment start (from workflow)
    data_end_date: 19990101T2300 #experiment current last timestep (from workflow)
    expver: a000 #experiment id
bridge:
    bridge_start_date: 19910101T0000 # could be null if no transfer has been done. first time step on the bridge
    bridge_end_date: 19920101T2300 # could be null if no transfer has been done. last time step on the bridge
    bridge_expver: o123 # could be null if no change has been done in the expver from hpc to bridge
```

possibly positioned in the `FDB_HOME` folder of the experiment. This should be written and updated by the workflow.

## Adaptation of the catalog generator

The catalog generator should be able to take a `FDB_HOME_DATABRIDGE` while the catalog source is generated.
If the workflow produces the yaml file generated above, this should be inserted instead of the string values with the dates that are inserted now by the catalog generator, meaning that the dynamical dates are turned on and that are now written in the catalog explicitely.

## Intake GSV update

Add to the intake gsv module the possibility to read a yaml file with the aspect described above and drop the reading from txt feature enabled atm, which is unused and deprecated.

oloapinivad:
## PR description:

Draft, more asap

Close #1617

----

 - [x] Changelog is updated.


oloapinivad:
Following the updates the `fdb_info_file` looks like this

```yaml
bridge:
  bridge_end_date: '19900102'
  bridge_end_hour: T2300
  bridge_start_date: '19900101'
  bridge_start_hour: T0000
  expver: t039
hpc:
  data_end_date: '19900102'
  data_end_hour: T2300
  data_start_date: '19900102'
  data_start_hour: T0000
  expver: t039
```

I updated the function that was reading introducing a static method that performs the validation. I am testing it now.

ainagaya:
Hi @oloapinivad ,
 
sorry I left this a bit aside due to other priorities, the easier implementation was the one that you saw with a `data_start_hour` field (due to the script structure) but I made some changes already last week to put them in the format that you requested in the first place (but I didn't have the time to re-run the experiment). 

But if the format with `data_start_hour` is fine for you, our script looks cleaner with it.

oloapinivad:
The format is fine, I wrote a small function that can deal with that. We still have to address some issues on our side, since the way dates are accounted for is a bit confusing at this stage.

oloapinivad:
@ainagaya so we had a chat and actually yes, there is the need of some modifications from your side we would like 😄 

We would like that the `hpc` block is renamed `data` and takes information on the ENTIRE start and end date of the simulation. this is the convention used inside AQUA. So something like this:

```yaml
bridge:
  bridge_end_date: '19900102'
  bridge_end_hour: T2300
  bridge_start_date: '19900101'
  bridge_start_hour: T0000
  expver: t039
data:
  data_end_date: '19900102' #this should be the end of the simulation
  data_end_hour: T2300
  data_start_date: '19900102' #this should be start of the simulation
  data_start_hour: T0000
  expver: t039 #this is the expver on the HPC
```

Of course if you can make the `hour` together with the `date` it would be easier but it is a minor thing.
Is this something that make sense? 
Thanks

ainagaya:
Hi @oloapinivad , 

Sure, no problem! 

This means that you always look for the data in the data bridge if it's available, right? 

What it means in workflow is that we won't update the file when the data is deleted from the HPC (wipe task), but only after the simulation runs (data update) and when it is transferred (bridge update).

Just FYI I'll put the hour in the same field as the hour (I had the changes done but they were not in the test file yet).

oloapinivad:
Exactly! in this way we gave priority to the data to the bridge, and we can also potentially deal with cases when data is on the bridge in the middle of the simulation

oloapinivad:
I included the most recent updates from @ainagaya so that everything is smoothly handled. 
@silviacaprioli could you check my changes to the catgen are ok?

silviacaprioli:
> I included the most recent updates from @ainagaya so that everything is smoothly handled. @silviacaprioli could you check my changes to the catgen are ok?

looks good to me!

mnurisso (approved):

mnurisso:
Merging

PR #1696, opened 2025-02-12
Fix net surface radiative flux and wind stresses in ecmean

jhardenberg:
## PR description:

All our DT ecmean plots are missing Net surface raditive flux and wind stresses. 
The reason is wrong variable names in the corresponding  ecmean interface file. This PR fixes this. 

----

 - [x] Changelog is updated.


oloapinivad:
Ah! well spotted, this is one of the problem @koldunovn  was mentioning today at the simulation meeting!

oloapinivad (approved):

oloapinivad:
BTW this is a very good case of gotfix that needs to go to the v013-operational as well. Indeed, the hotfix should have been branched from the operationa and then ported to the main (so this PR is somehow "wrong" because need a second twin PR and we cannot port this changes to the operational). We could try to cherry pick commits, which should be probably the way to go.  

We should also start producing two different aqua explorer as originally discussed , one with the v013-operational branch and one with the development as discussed in the past.

jhardenberg:
Good point about the hotfix to operational. I will open a second PR for that (in this case it is easy, it is just one file which was modified).  The problem will be for more serious changes where indeed it will be better to start from operational and only later make a PR to the main. 

As for the explorer, yes, following it, even if at the moment we have another problem (boto3 not pushing to lumi-o anymore).

mnurisso:
Merging

PR #1697, opened 2025-02-13
[OP] Fix net surface radiative flux and wind stresses in ecmean 

oloapinivad:
## PR description:

Cherry picking commit from #1696

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 Unreleased in the current development version:
 
+AQUA diagnostics complete list:
+- Fix net surface radiative flux and wind stresses in ecmean (#1697)
+
 ## [v0.13.1]
 
 Main changes are:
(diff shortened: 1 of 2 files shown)


jhardenberg:
maybe still add changelog?

oloapinivad:
> maybe still add changelog?

yes but this has to be different in the two PRs

jhardenberg:
In the future we should also think about a policy so that the PR titles and the entry in the changelog are similar, so that it is easy to understand that the two PR are paired (maybe with the operational PR having "(operational) " in the title. So I would have named this PR "HotFix net surface radiative flux and wind stresses in ecmean (operational)" maybe?

oloapinivad:
Maybe [Operational] at the very beginning is better?

jhardenberg:
One could claim that labels could be enough, but labels come and go easily.  I would indeed include something in the title to make sure that nobody gets confused between the two. Maybe shorter "[Op]" at the beginning ?

jhardenberg (approved):
looks good. Merge?

oloapinivad:
I think this can be merged now

mnurisso:
Merging

PR #1704, opened 2025-02-14
Hotfix: push_s3 compatibility with boto3>=1.36.0

jhardenberg:
## PR description:

This hotfix restores compatibility with boto3 >= 1.36.0.

`push_s3.py` had stopped working after recent environment updates, throwing this error message:
```
Error uploading file: Failed to upload /users/jvonhar/test.txt to aquatest2/test.txt: An error occurred (XAmzContentSHA256Mismatch) when c
alling the PutObject operation: None
```

The problem started with boto3 version >= 1.36.0, earlier versions worked.
This solution is better than a pin, since it fixes the issue both with recent and older versions of boto3.

We will probably need to fix this also in 0.13-operational if we plan to push future op figures to the dashboard.

----

 - [x] Changelog is updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Unreleased in the current development version (target v0.14):
 
 AQUA core complete list:
 
+- push_s3 compatibility with boto3>=1.36.0 (#1704)
 - Rsync option for push_analysis.sh (#1689)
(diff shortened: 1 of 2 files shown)


jhardenberg:
@mcadau this is the solution which you suggested. Does all look good?

mcadau (approved):

mnurisso:
I'm merging.

@jhardenberg or @mcadau could you duplicate the PR for the [OP] release?

PR #1706, opened 2025-02-14
Workflow changes to aqua push

ainagaya:

I didn't open an issue, but this were the changed that I had to do to push within the workflow (expid a26z)

----

 - [ ] Changelog is updated.

### DIFF
--- a/cli/aqua-web/push_analysis.sh
+++ b/cli/aqua-web/push_analysis.sh
@@ -216,9 +216,17 @@ fi
 
 cd $repo
 if [ $update -eq 1 ]; then
-    git checkout $branch
-    git pull
+    if git show-ref --verify --quiet refs/heads/$branch; then
+        # Branch exists locally
+        git checkout $branch
+        git pull origin $branch
+    else
+        # Branch does not exist locally
+        git checkout -b $branch
+        git pull origin $branch || echo "No remote branch to pull from"
+    fi
 fi
+
 
 echo "Updated figures in bucket $bucket"  > updated.txt
 echo "on $(date) for the following experiments:" >> updated.txt
@@ -251,7 +259,7 @@ if [ -f "$exps" ]; then
     done < "$exps"
 else  # Otherwise, use the second argument as the experiment folder
     log_message INFO "Collect figures for $exps and converting to png"
-    collect_figures "$indir" "$exps" $wipe
+    collect_figures "$indir" "$exps"
     convert_pdf_to_png "$exps"
     make_contents "$exps" "$config" # create catalog.yaml and catalog.json
     push_lumio $bucket "$exps" "$rsync"
@@ -268,7 +276,7 @@ if [ $update -eq 1 ]; then
     if [ "$dry" -eq 1 ]; then
         log_message INFO "Dry run, not pushing to the repository"
     else
-        git push
+        git push origin $branch
         log_message INFO "Pushed new figures to lumi-o"
     fi
 fi


mnurisso:
Hi @ainagaya after Jost review we will have to duplicate the PR, since the changes I imagine will be both needed in the future releases and as hotfix of the 0.13.1.

jhardenberg:
Hi @ainagaya , after recent changes (which are documented in the 'dashboard') section of the documentation - and see also the message which I left in the dashboard channel on Mattermost),  aqua-web does not rely anymore on figures stored on github. Instead the figures are now stored on lumi-o and retrieved by aqua-web.
So now there is not much which needs to change on the aqua-web githib repo itself, the only thing which we change when we push is the `update.txt` file in the root used to record the latest change and to trigger rebuilding of the web site by OpenShift. 
This means that the entire idea that from the workflow you would push to separate branches is not meaningful anymore
I see that here you reintroduce the creation of branches ... but I do not see why that would be needed.
If you want we could organize a moment to have a short meeting next week so that we aling on these issues, what do you think ?

ainagaya:
Hi @jhardenberg , thanks.

Do you have an example of how to call that script operationally?

Right now we are using:

```
bash ${LOCAL_AQUA}/cli/aqua-web/push_analysis.sh --branch ${branch} --no-convert ${indir} ${exps}
```

But from the documentation I see

```
bash push-analysis.sh [OPTIONS] INDIR EXPS
```

Do you want the workflow to use any specific flags?

I also see that we need AWS credentials. Are there instructions to set them up somewhere? Thanks :)

jhardenberg:
Indeed now you should be using

```
cd ${LOCAL_AQUA}/cli/aqua-web 
bash push_analysis.sh  ${indir} $explist
```
or 
```
cd ${LOCAL_AQUA}/cli/aqua-web 
bash /push_analysis.sh  ${indir} $explist`
```

if you have a file listing the experiments to be pushed in the format
`catalog model experiment`

As discussed on mattermost, if you are not interested in updating immediately aqua-web, you could also use the option `-d` which prevents the push to aqua-web and wait for the next update triggered by our workflow the next morning.

In any case (**important**):
*  the option `--branch` still exists, but it will be removed soon, it has no reason to be or to be used anymore.
* the option "--no-convert` also should  not be used: if you do not convert figures to png locally, then actually they will not be displayed, because aqua-web is not doing this anymore. This is for compatibility with the new dashboard which needs png directly.  We may keep the option because we hope that in the future all figures are produced directly also as png, no need to convert them.  The issue was that you do not have imagemagick on the local vm? Maybe a container could help?

As for permissions to write to lumi-o: either you store a token for lumi-o AWS credentials (as described in the docs), but this would need to be renewed every 7 days, or we can give writing access to specific IPs (if your workflow VM has a fixed IP or IP range, we could add it to the allowed IPs)

@ainagaya if you prefer we could organize a short meeting in the next days to align on all this.

jhardenberg:
> I also see that we need AWS credentials. Are there instructions to set them up somewhere? Thanks :)

The docs (in the dashboard section) describe how to save the AWS credentials.
The credentials themselves can be retrieved from https://docs.lumi-supercomputer.eu/storage/lumio/auth-lumidata-eu/

ainagaya:
Thanks @jhardenberg , we updated the workflow accordingly. I opened a PR https://github.com/DestinE-Climate-DT/AQUA/pull/1727 to fix a small issue. @sebastianbeyer is also helping with the testing.

PR #1703, opened 2025-02-13
Template parser

mnurisso:
## PR description:

A common parser and a shared open/close cluster function is introduced

## Issues closed by this pull request:

Close #1612 

----

Please leave the checkboxes that apply to this pull request.
If you find a missing checkbox, please add it to the list.
Make sure to complete the checkboxes before applying the "ready to merge" label.
Please apply the "run tests" label if you want to trigger CI tests.

 - [ ] Tests are included if a new feature is included.
 - [ ] Documentation is included if a new feature is included.
 - [ ] Docstrings are updated if needed.
 - [ ] Changelog is updated.
 - [ ] Notebooks which requires changes are updated. 
 - [ ] environment.yml and pyproject.toml are updated if needed, together with the lumi installation tool. 
 - [ ] Diagnostic config files path are updated in the console if needed

PR #1392, opened 2024-10-16
ETCCDI indices

mnurisso:
## PR description:

Implementation of CLI scripts to evaluate ETCCDI indices as described [here](https://etccdi.pacificclimate.org/list_27_indices.shtml).
Only a preliminar list of indices is available.

They work at the moment with native healpix data and no regrid, but this can be easily extended.

A more rigorous structure as a class, with definition yaml files and only a final cli to run as a proper diagnostic would be beneficial.

----

Please leave the checkboxes that apply to this pull request.
If you find a missing checkbox, please add it to the list.
Make sure to complete the checkboxes before applying the "ready to merge" label.
Please apply the "run tests" label if you want to trigger CI tests.

 - [ ] Tests are included if a new feature is included.
 - [ ] Documentation is included if a new feature is included.
 - [ ] Docstrings are updated if needed.
 - [ ] Changelog is updated.
 - [ ] Notebooks which requires changes are updated. 
 - [ ] environment.yml and pyproject.toml are updated if needed, together with the lumi installation tool. 
 - [ ] Diagnostic config files path are updated in the console if needed

PR #1710, opened 2025-02-15
ETCCDI (SDII template for other indices)

mnurisso:
## PR description:

The PR propose a new implementation of #1392 with the implementation of the full pipeline designed in #1682.
A first core class in common to all the ETCCDI indices derived from the daily statistics is presented.
Then a specific SDII class taking care of the daily data is introduced, with the role of saving the netcdf data of a single model.
Finally a plot class has the role of saving the pdf/png data.

A CLI working with the core parser introduced in #1703 is developed and has been tested.
The config file are installed when aqua install is called.

Notice that the plot routine will work only with healpix data since they are using radically different plot packages.

@jhardenberg I'm more than happy to change form ETCCDI to ET-SCI, but I found only few info on them, maybe we could discuss online this.

I'd like instead here doubt or suggestion in the actual implementation and also what would be the minimal requirement to merge this (finish docs for just this single index is quick, adding more indices radically different may require time).

This have been tested on healpix 10 hourly data to check with the highest resolution available.

----

 - [ ] Changelog is updated.


mnurisso:
We should investigate the xclim package.
Here a snippet of a possible quick test to expand:

```python
from aqua import Reader
import xclim
import healpy as hp

reader = Reader(model="IFS-NEMO", exp="historical-1990", source="hourly-hpz7-atm2d")
data = reader.retrieve(var="tprate")
pr= data["tprate"].sel(time=slice("2000-01-01", "2000-12-31T23:00"))
prd = pr.aqua.timmean(freq='D')
prdl = prd.compute()  # This only for memory issues in the notebook

sdii = xclim.indicators.icclim.SDII(pr=prdl, freq="YS")

hp.mollview(sdii.squeeze(), title="SDII", nest='nest')
```

oloapinivad:
Closing as deprecated

PR #1711, opened 2025-02-17
LRA capacity to deal with regional selection

oloapinivad:
## PR description:

There are situations were building a limited area LRA might be useful. This PR tries to implement this functionality, which requires a bit of hacking of filenames

 - [x] Changelog is updated.


src/aqua/cli/lra.py
```diff
@@ -115,11 +119,13 @@ def lra_execute(args):
     lra_cli(args=args, config=config, catalog=catalog, resolution=resolution,
             frequency=frequency, fix=fix,
             outdir=outdir, tmpdir=tmpdir, loglevel=loglevel,
+            region=region,
             definitive=definitive, overwrite=overwrite, rebuild=rebuild,
             default_workers=default_workers,
             monitoring=monitoring, do_zarr=do_zarr, verify_zarr=verify_zarr, only_catalog=only_catalog)
 
 def lra_cli(args, config, catalog, resolution, frequency, fix, outdir, tmpdir, loglevel,
+            region=None,
```
mnurisso, line 128: we're missing docstrings here

src/aqua/lra_generator/lra_generator.py
```diff
@@ -65,6 +68,8 @@ def __init__(self,
                                      are found
             nproc (int, opt):        Number of processors to use. default is 1
             loglevel (string, opt):  Logging level
+            region (dict, opt):    Region to be processed, default is None,
```
mnurisso, line 71: From docstring, it is not clear the format of the dictionary expected

src/aqua/lra_generator/lra_generator.py
```diff
@@ -144,14 +149,24 @@ def __init__(self,
             raise KeyError('Please specify resolution.')
         self.logger.info('Variable(s) to be processed: %s', self.var)
 
+        if region is not None:
+            self.logger.info('Region to be processed: %s', region)
```
mnurisso, line 153: Are you going to print a dictionary here, it is the expected behavior?

src/aqua/lra_generator/lra_generator.py
```diff
@@ -580,6 +609,9 @@ def _write_var_catalog(self, var):
         self.logger.info('Processing variable %s...', var)
         temp_data = self.data[var]
 
+        if self.region:
```
mnurisso, line 618: A philosophical question. If you leave the area_selection here it means the original data mush have lon/lat as coordinates.
If you move it below after the regridding it may be not as accurate, but it will work in any case.

src/aqua/lra_generator/lra_generator.py
```diff
@@ -144,14 +149,24 @@ def __init__(self,
             raise KeyError('Please specify resolution.')
         self.logger.info('Variable(s) to be processed: %s', self.var)
 
+        if region is not None:
+            self.logger.info('Region to be processed: %s', region)
+            self.region = region
+            if self.region['lon'] is None:
```
mnurisso, line 155: Are we sure that lon and lat have to be not None? What if I want a tropical band for example? One of the two None is fine, while both None should be a raise KeyError

templates/lra/config_lra.tmpl
```diff
@@ -5,6 +5,9 @@ target:
   resolution: r100
   frequency: monthly
   catalog: {{ catalog }}
+  #region:
+  #  name: Italy
+  #  boundaries: [35, 50, 6, 18]
```
mnurisso, line 10: Not reflecting the dictionary used in the code

mnurisso (commented):
I wanted to give a quick check but I left more comment than expected 😶‍🌫️ 
The real doubt is only on where to perform the area selection

oloapinivad:
A lot of changes included, this seems ready to me

mnurisso:
> A lot of changes included, this seems ready to me

Do we care about the test percentage drop or shall I merge?

oloapinivad:
I think we can ignore this, I am going to reintroduce the codecov file in #1797 so it should be ok

mnurisso (approved):

mnurisso:
Ok, let's merge it

PR #1712, opened 2025-02-17
Timeseries refactor

mnurisso:
## PR description:

Timeseries refactor according to #1682 

----

 - [x] Tests adapted
 - [ ] Documentation adapted
 - [x] Changelog is updated.
 - [x] PlotTimeseries class
 - [x] CLI refactor


src/aqua/util/time.py
```diff
@@ -21,6 +21,7 @@ def frequency_string_to_pandas(freq):
         'daily': 'D',
         'weekly': 'W',
         'monthly': 'MS', #this implies start of the month
+        'annual': 'YS', #this implies start of the year
```
mnurisso, line 24: Is this ok? @oloapinivad

oloapinivad: yes is the same yearly below

mnurisso:
So far we have a working class for the timeseries of a single model, together with a CLI that can produce all the netcdf for the models and references. A PlotTimeseries class is still missing due to a necessary decision on how to proceed with the other diagnostics and the plot_timeseries refactor.

I'll continue with the SeasonalCycle in order to produce for it as well all the necessary netcdf.

src/aqua_diagnostics/timeseries/base.py
```diff
@@ -0,0 +1,268 @@
...
+from aqua.exceptions import NoDataError
+from aqua.logger import log_configure
+from aqua.util import ConfigPath
+from aqua.util import frequency_string_to_pandas
+from aqua.util import load_yaml, eval_formula, convert_units
+from aqua.diagnostics.core import Diagnostic, start_end_dates
+
+
+class BaseMixin(Diagnostic):
+    """Region selection mixin class. Used by """
```
mnurisso, line 13: what did I wrote here?

mnurisso:
Only documentation is missing. 
@jhardenberg it would be great to test this on any run, do you think it can be done on HPC2020 or is it too much effort? Otherwise we can wait for new resources in the project.

oloapinivad (approved):
This has been tested, it works, and it is too much for a detailed review, let's merge!

PR #1726, opened 2025-02-19
LRA zip compression for yearly files

oloapinivad:
## PR description:

Enable LRA compression also for concat yearly files, which was missing. I would not applicate this to operational. 

 - [x] Changelog is updated.


mnurisso (approved):

mnurisso:
Merging

PR #1728, opened 2025-02-19
[OP] Refactor GSV retrieval logic for Dask compatibility

### LINKED ISSUE #1715: Infinite hang when retrieving data from the bridge in IFS-FESOM storylines
### machine

Lumi

### catalog

climatedt-phase1

### version

main

### What happened?

I am experiencing an infinite hang when I try to access the data from the storylines, irrespective of which experiments. 
This holds true only for data on the bridge, while more recente data from 2024 on the HPC can be accessed. Please notice that lazy access works as it should, and the problem shows up only when asking to access the underlying arrays.
Interestingly, the problem is not there if we run with container v0.13.1. 

```python
from aqua import Reader
reader = Reader(model='IFS-FESOM', exp='story-2017-historical', source='hourly-hpz7-atm2d', regrid='r100', loglevel='debug')
data = reader.retrieve(startdate='2022-03-01', enddate='2022-03-02')
data.isel(time=0)['2t'].values
```

### Error messages

Infinite hang

Yes

mnurisso:
## PR description:

In order to fix #1715 we introduce an internal check that initialize a single GSV if we're not in a cluster, otherwise one instance per worker is created

Close #1715 

----

 - [ ] Changelog is updated.

### DIFF
--- a/src/aqua/gsv/intake_gsv.py
+++ b/src/aqua/gsv/intake_gsv.py
@@ -6,6 +6,7 @@ import eccodes
 import xarray as xr
 import numpy as np
 import dask
+from dask.distributed import get_worker
 from ruamel.yaml import YAML
 from aqua.util.eccodes import get_eccodes_attr
 from aqua.util import to_list
@@ -31,6 +32,7 @@ class GSVSource(base.DataSource):
     name = 'gsv'
     version = '0.0.2'
     partition_access = True
+    gsv = None # This is necessary to be shared or not depending on the dask cluster or not
 
     _ds = None  # _ds and _da will contain samples of the data for dask access
     _da = None
@@ -481,7 +483,7 @@ class GSVSource(base.DataSource):
 
         # this is needed here and not in init because each worker spawns a new environment
         gsv_log_level = _check_loglevel(self.logger.getEffectiveLevel())
-        gsv = GSVRetriever(logging_level=gsv_log_level)
+        gsv = self.get_gsv(gsv_log_level)
 
         self.logger.debug('Request %s', request)
         dataset = gsv.request_data(request, use_stream_iterator=fstream_iterator, 
@@ -641,6 +643,19 @@ class GSVSource(base.DataSource):
             self.logger.info('Automatic FDB date range: %s - %s', start_date, end_date)
 
         return start_date, end_date
+
+    def get_gsv(self, gsv_log_level):
+        if "DASK_WORKER" in os.environ:
+            # Running inside a Dask worker → create a new instance per worker
+            worker = get_worker()  # Get current worker
+            if not hasattr(worker, "gsv"):
+                worker.gsv = GSVRetriever(logging_level=gsv_log_level)
+            return worker.gsv
+        else:
+            # Running in serial → use a single shared instance
+            if GSVSource.gsv is None:
+                GSVSource.gsv = GSVRetriever(logging_level=gsv_log_level)
+            return GSVSource.gsv
 
 
 # This function is repeated here in order not to create a cross dependency between GSVSource and AQUA


jhardenberg:
Please check my branch `gsv-fix-alt`: it has a solution along these lines  but much simpler because it does not need to check if running in cluster mode or not. I tested it both in no cluster and in cluste rmodes and it seems to work well. To fully test, notice that it is better to do:
` data.isel(time=slice(0,25))['2t'].values` (so that more than one chunk is read).

Basically it has just these lines of code in `_get_partition()`:

```
        if not hasattr(GSVSource, 'gsv') or not GSVSource.gsv:
            GSVSource.gsv = GSVRetriever(logging_level=gsv_log_level)

        self.logger.debug('Request %s', request)
        dataset = GSVSource.gsv.request_data(request, use_stream_iterator=fstream_iterator, 
                                   process_derived_variables=False) #following 2.9.2 we avoid derived variables
```

jhardenberg:
Notice that for some mysterious reason storing the gsv object in `GSVSource` works, while storing in self does not!

src/aqua/gsv/intake_gsv.py
```diff
@@ -642,6 +644,19 @@ def parse_fdb(self, start_date, end_date):
...
+    def get_gsv(self, gsv_log_level):
+        if "DASK_WORKER" in os.environ:
+            # Running inside a Dask worker → create a new instance per worker
+            worker = get_worker()  # Get current worker
+            if not hasattr(worker, "gsv"):
+                worker.gsv = GSVRetriever(logging_level=gsv_log_level)
+            return worker.gsv
+        else:
+            # Running in serial → use a single shared instance
+            if GSVSource.gsv is None:
```
oloapinivad, line 656: uhm, what is `GSVsource` here?

oloapinivad: we are importing `from gsv.retriever import GSVRetriever` so I am wondering how do you get that object

jhardenberg: it is the class itself, defined above at line 29.

oloapinivad: so why using the class and not `self` here?

mnurisso: Because from my test I didn't find a working solution with self, we can check on Jost branch

oloapinivad: this is quite strange. I am wondering if in this way you are "redefining" the class itself.

jhardenberg: No, I confirm, self does not work.

jhardenberg:
@mnurisso plz give a serious look at my version in gsv-fix-alt, it follows the same idea as here but it has the advantage that there is less code, no need  to check for cluster/no cluster and above all, when run in cluster mode it initializes GSVRetriever only once per worker, reinitializing only if the worker happens to die. Overall if we confirm that all this works it is much better both in cluster and non/cluster mode, since we initialize the GSVRetrieve only when strictly necessary (before we were initializing at every retriev, something which apparently pyfdb has begun to dislike when accessing the bridge)

src/aqua/gsv/intake_gsv.py
jhardenberg: Whe dask runs in cluster mode it does something strange with classes: each worker reinitializes the class but apparently dies not properly run __init__. self is not  preserved and needs to be passed between workers by pickling its contents. Problem is that a GSVRetriever object is unpickleable.

mnurisso:
> @mnurisso plz give a serious look at my version in gsv-fix-alt, it follows the same idea as here but it has the advantage that there is less code, no need to check for cluster/no cluster and above all, when run in cluster mode it initializes GSVRetriever only once per worker, reinitializing only if the worker happens to die. Overall if we confirm that all this works it is much better both in cluster and non/cluster mode, since we initialize the GSVRetrieve only when strictly necessary (before we were initializing at every retriev, something which apparently pyfdb has begun to dislike when accessing the bridge)

I'm first fixing tests for the v0.13-operational and then testing both the branches, sure

PR #1733, opened 2025-02-21
GSV intake driver instantiates GSVRetriever object only if needed

jhardenberg:
## PR description:

This modifies the GSV intake driver so that a new `GSVRetriever` object is reinitialized only if needed (bridge/hpc switch, eccodes switch).

For main only, no OP.

Connected to #1715
Followup to #1728

----

 - [x] Changelog is updated.


src/aqua/gsv/intake_gsv.py
```diff
@@ -477,14 +484,25 @@ def _get_partition(self, ii, var=None, first=False, onelevel=False):
...
         # this is needed here and not in init because each worker spawns a new environment
         gsv_log_level = _check_loglevel(self.logger.getEffectiveLevel())
-        gsv = GSVRetriever(logging_level=gsv_log_level)
+
+        if flag_eccodes_switch or not hasattr(GSVSource, 'gsv') or not GSVSource.gsv or rebuild_gsv:
+            GSVSource.gsv = GSVRetriever(logging_level=gsv_log_level)
 
         self.logger.debug('Request %s', request)
-        dataset = gsv.request_data(request, use_stream_iterator=fstream_iterator, 
+        dataset = GSVSource.gsv.request_data(request, use_stream_iterator=fstream_iterator,
```
oloapinivad, line 505: I am not sure we still need the `fstream_iterator` distinction, it seems to work in both cases if it is true. 
Also, could you explain once for all why you cannot use `self.gsv` instead of `GSVSource.gsv` here?

src/aqua/gsv/intake_gsv.py
```diff
@@ -477,14 +484,25 @@ def _get_partition(self, ii, var=None, first=False, onelevel=False):
...
+            self.logger.debug("FDB5_CONFIG_FILE has been changed to from %s to %s", current_fdb_path, os.environ.get("FDB5_CONFIG_FILE", None))
+            rebuild_gsv = True
+
+        flag_eccodes_switch = self._switch_eccodes()
 
         # this is needed here and not in init because each worker spawns a new environment
         gsv_log_level = _check_loglevel(self.logger.getEffectiveLevel())
-        gsv = GSVRetriever(logging_level=gsv_log_level)
+
+        if flag_eccodes_switch or not hasattr(GSVSource, 'gsv') or not GSVSource.gsv or rebuild_gsv:
```
mnurisso, line 501: For easy further development I'd add here the idea behind this minimization of the GSV instances, if we have to come back to this code otherwise it can be tricky to remember why we did this.

src/aqua/gsv/intake_gsv.py
oloapinivad: as it is defined such `GSVsource` definition looks like a recursive structure. Despite I trust you, I am still puzzled it can work.

mnurisso: It's not recursive, `gsv` is defined globally and not in the `__init__()` the call for the dataset uses the `gsv`, which may have been defined in another `GSVSource` instance. However, since we did careful check on when the `GSVRetriever` should be redefined, we never use `gsv` with a wrong FDB for which it was not instantiated the `GSVRetriever` class

oloapinivad: I don't understand your point. I am not talking of `GSVretriever` here, rather of `GSVsource`, which is the name of the overall class. At L505 you are accessing the `GSVsource.gsv` from inside the class itself, so you are creating a new external `GSVsource.gsv` object rather the one of the class it self. Why do you need to call it `GSVsource` and not use another random variable at that point?

mnurisso: No, I'm accessing `GSVSource` as class and checking if the GLOBAL `GSVSource` has the `GSVSource.gsv`. No `__init__` is done, since you don't use any `__init__` we you try to access an attribute.

So there is no recursion, the only "strange" usage is the check on the global, which can take a `gsv` attribute that has been generated in another `GSVSource` instance.

oloapinivad: Your interpretation is not far from what is happening: so I have just discovered the difference between a class attributes (`GSVSource.gsv`), shared among all the instances, and a instance attributes (`self.gsv`).

A dangerous downside that I might see is that two different `Reader` might share the save `gsv`. However, in principle, if you are pointing to a different GSV you should change the `FDB_HOME` so you create a new object.

mnurisso: Is not a downside, if they share the same `gsv` is because the `FDB_HOME` is the same and `eccodes` as well so you're absolutely saving time. 

In my test, from test 3 to test 4 you change source but not `FDB_HOME`. So a new `Reader` instance does not create a new `GSVSource.gsv` and it perfectly works since you're accessing the same FDB.

src/aqua/gsv/intake_gsv.py
```diff
@@ -372,11 +372,16 @@ def _get_schema(self):
         return schema
 
     def _switch_eccodes(self):
+        """
+        Internal method to switch ECCODES version if needed
```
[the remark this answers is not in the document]
jhardenberg: Lol, I knew I should not have added it ;) 
Will do

jhardenberg:
@mnurisso any tests we need to add?

mnurisso:
> @mnurisso any tests we need to add?

We have only one FDB on tests so I find difficult to understand how to test this, for the moment I think that's fine

mnurisso:
After tests I'm merging this

oloapinivad:
Tests passed, merging

oloapinivad (approved):

PR #1738, opened 2025-02-22
Further intake-gsv fixes

jhardenberg:
## PR description:

This PR is only to start working and discussing on a solution towards the continued problems discussed in #1715 

In this branch I keep the previous structure which instantiates a new GSVRetriever object the first time or every time `FDB_HOME` changes, but additionally it also instantiates a new object anyway  (as in older versions of AQUA) before running the retrieve. This means that the first time this is run two gsv objects are instantiated and again, two are instantiated if `FDB_HOME` changes. 
Apparently this allows the code to run again correctly both on bridge and hpc and avoids the 'infinite hang' issue. Basically it implements the first solution discussed at the top of #1715 where it was found that instantiating GSVRetriever twice seemed to solve the problem.  Of course the issue that accessing in sequence bridge, hpc, bridge crashes remains.

Another change here is that now the extra 'initial' GSVRetriever object is temporarily stored in `self` (not `GSVSource`) but then actually not used. Here the true magic begins: if instead of storing this extra object in `self` I discard it, writing `_ = GSVRetriever(logging_level=gsv_log_level)` in line 556 .... all this stops working!  
Clearly all this is a hack around some crazy bug of pyfdb or fdb5. 

In any case, please check, it seems to me that this branch at least avoids the infinite hang and works again.

 - [x] Changelog is updated.

### DIFF
--- a/src/aqua/gsv/intake_gsv.py
+++ b/src/aqua/gsv/intake_gsv.py
@@ -237,6 +237,8 @@ class GSVSource(base.DataSource):
         else:
             self.chunking_vertical = None  # no vertical chunking
 
+        self.gsv = None
+
         self._switch_eccodes()
 
         super(GSVSource, self).__init__(metadata=metadata)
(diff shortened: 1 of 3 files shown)


src/aqua/gsv/intake_gsv.py
```diff
@@ -549,13 +551,14 @@ def _get_partition(self, ii, var=None, first=False, onelevel=False):
         gsv_log_level = _check_loglevel(self.logger.getEffectiveLevel())
 
         # The following is done to recycle the GSVRetriever instance: it has to be changed only if FDB_HOME or ECCODES changed
-        # Notice that this is done with a class variable, so it is shared among all instances of GSVSource
-        if flag_eccodes_switch or not hasattr(GSVSource, 'gsv') or not GSVSource.gsv or rebuild_gsv:
-            GSVSource.gsv = GSVRetriever(logging_level=gsv_log_level)
-
+        if flag_eccodes_switch or rebuild_gsv or not hasattr(self, 'gsv') or not self.gsv:
```
mnurisso, line 554: Here I think with the modifications the third condition is never met and can be removed

mnurisso (approved):
I'm ok with the PR, but we need to comment in the code mentioning the issue number so that in a further development we are able to remember what did happen here and why there is a double GSV without any apparent motivation

src/aqua/gsv/intake_gsv.py
jhardenberg: Actually ... not. In cluster mode the workers see only what is passed to them by `__setstate__()` and `__getstate__()`. So since currently `gsv` is not included, the worker's self will not containe the `self.gsv = None` initialized in main. We could include this in `__setstate__()` but since a GSVRetriever object cannot be pickled, it seems actually better to leave this `hasattr` check here and rather remove the now useless `self.gsv = None`  in `__init__()`. the latest commit does this

mnurisso: Ok, make kinda sense, still a bit obscure 🤣

jhardenberg:
I added a test for bridge sources, if https://github.com/DestinE-Climate-DT/Climate-DT-catalog/pull/134 is merged we can run it.

mnurisso:
> I added a test for bridge sources, if [DestinE-Climate-DT/Climate-DT-catalog#134](https://github.com/DestinE-Climate-DT/Climate-DT-catalog/pull/134) is merged we can run it.

It is merged

jhardenberg:
Ok, all looks good, I think we can merge

mnurisso (approved):
Perfect, let's merge

PR #1748, opened 2025-02-24
More flexible action including also Dependabot case

oloapinivad:
## PR description:

This  - in theory - should allow for more precise label-triggering actions as well as action to be triggered when dependenabot is set up.


oloapinivad:
If this works, can be a nice solution for dependabot, what do you think @mnurisso?

mnurisso (approved):

mnurisso:
I agree, merged

PR #1758, opened 2025-02-25
remove worthless blocks from coverage

oloapinivad:
## PR description:

Remove some files from coverage count. @mnurisso anything else that is worth to be ignored?


mnurisso:
I think the `AQUA/cli` folder is already ignored is it correct? If so, that's fine for me.

mnurisso (approved):

PR #1763, opened 2025-02-26
retrieve method for Diagnostic class can return data

mnurisso:
## PR description:

Sometimes it can be useful to return data from a retrieve instead of updating the self.data
This PR introduces such update for the Diagnostic primitive class.

 - [x] Changelog is updated.


mnurisso:
We will introduce a `_retrieve` function as static_method instead of the actual implementation and then the static method is internally called by the retrieve actual function

oloapinivad (approved):
It seems an useful extension, we will see in the near future if some adjustment are needed

PR #1782, opened 2025-03-01
Installation of AQUA on ECMWF HPC2020

jhardenberg:
This PR adds documentation and an installation script  to use efficiently AQUA on HPC2020 at ECMWF.

 - [x] Changelog is updated.

### DIFF
--- a/cli/hpc2020-install/hpc2020_install.sh
+++ b/cli/hpc2020-install/hpc2020_install.sh
@@ -141,5 +141,5 @@ while true; do
   esac
 done
 
-log_message WARNING "AQUA environment has been installed, please remember to run 'aqua install hpc2020' and 'aqua add hpc2020 -e <path/to/your/catalogue>'"
-log_message WARNING "hpc2020 is the name of this machine in AQUA syntax. You will need to provide a catalogue for hpc2020 in order to use the diagnostics."
+log_message WARNING "AQUA environment has been installed, please remember to run 'aqua install hpc2020' and 'aqua add <your-catalogue-name> -e <path/to/your/catalogue>'"
+log_message WARNING "hpc2020 is the name of this machine in AQUA syntax. You will need to provide a catalogue for experiments on hpc2020 in order to use the diagnostics."
--- a/cli/hpc2020-install/hpc2020_install.sh
+++ b/cli/hpc2020-install/hpc2020_install.sh
@@ -132,7 +132,7 @@ while true; do
       ;;
     [Nn])
       log_message WARNING "AQUA not added to PATH in .bash_profile"
-      log_message $next_level_msg_type "You can use AQUA by loading the tykky module (module load tykky) and then activating the environment (tykky activate aqua)"
+      log_message $next_level_msg_type "You can use AQUA by loading the tykky module 'module load tykky' and then activating the environment 'tykky activate aqua'"
       break
       ;;
     *)
@@ -141,4 +141,4 @@ while true; do
   esac
 done
 
-log_message WARNING "AQUA environment has been installed, please remember to to run 'aqua install' and 'aqua add lumi'"
+log_message WARNING "AQUA environment has been installed, please remember to run 'aqua install' and 'aqua add hpc2020' (hpc2020 is the name of this machine in AQUA syntax)"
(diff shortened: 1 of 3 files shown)


jhardenberg:
Just a remark: it is actually essential to follow this procedure to install AQUA on HPC2020. Trying to use it directly using conda or mamba results in extremely slow performance: a simple test python script importing AQUA, opening a dataset and retrieving the data, may take up to 15 minutes with mamba, while using the containerized approach described her it takes just 8 seconds.

jhardenberg:
The process is actually quite simple, it only gets complicated by details like paths to $AQUA and the bug described in the docs. I am writing actually a `hpc2020_install.sh` to make everything simpler.

jhardenberg:
I added an installation script in `cli/hpc2020-install` and updated the documentation. 

Please check, but this looks good to me and I think that it could be merged. 

The recommended way to use aqua is then to do `module load tykky` and then use a 'conda-like' syntax : `tykky activate aqua` .

mnurisso (approved):

mnurisso:
I'm testing this and then merge. @jhardenberg do you have a preference about the machine name that we should use? atos or hpc2020?

cli/hpc2020-install/hpc2020_install.sh
```diff
@@ -0,0 +1,144 @@
...
+      log_message $next_level_msg_type "You can use AQUA by loading the tykky module (module load tykky) and then activating the environment (tykky activate aqua)"
+      break
+      ;;
+    *)
+      log_message ERROR "Invalid response. Please enter 'y' or 'n'."
+      ;;
+  esac
+done
+
+log_message WARNING "AQUA environment has been installed, please remember to to run 'aqua install' and 'aqua add lumi'"
```
mnurisso, line 144: I'd remove aqua add lumi here :)

I'd add a line stating, in AQUA syntax this machine is called: 'name-that-we-decide'

mnurisso (approved):
I tested on atos as well, if we fix the typo in the message at the end of the installation for me we can merge it.

jhardenberg:
> I'm testing this and then merge. @jhardenberg do you have a preference about the machine name that we should use? atos or hpc2020?

We often call it 'atos' but ECMWF seems to officially call the machine HPC2020. Atos is the name of the company which sold and installed the machine https://en.wikipedia.org/wiki/Atos
So I would go with HPC2020

mnurisso:
Ok, @jhardenberg I'm merging this. I'd like to follow up with the aqua add option from another repo than the climatedt catalog one, also aqua add hpc2020 is currently misleading since there's no hpc2020 catalog, but it's good to mention which is the supported machine name

jhardenberg:
Should we change the final message so that aqua add from a different catalog is suggested ? It is true that there is no 'standard' hpc2020 catalogue (in general `aqua add <catalog>` refers to ClimateDT catalogues

mnurisso:
> Should we change the final message so that aqua add from a different catalog is suggested ? It is true that there is no 'standard' hpc2020 catalogue (in general `aqua add <catalog>` refers to ClimateDT catalogues

Maybe aqua add obs or aqua add ci could be an option.
Notice also that now there's no way to guide the user to dynamically set up the grids, areas, weights paths so that the deployment in a not climatedt machine it's still tricky.

We can think about another aqua console command to set these paths if you're in a new machine

cli/hpc2020-install/hpc2020_install.sh
```diff
@@ -0,0 +1,145 @@
...
+      log_message $next_level_msg_type "You can use AQUA by loading the tykky module 'module load tykky' and then activating the environment 'tykky activate aqua'"
+      break
+      ;;
+    *)
+      log_message ERROR "Invalid response. Please enter 'y' or 'n'."
+      ;;
+  esac
+done
+
+log_message WARNING "AQUA environment has been installed, please remember to run 'aqua install hpc2020' and 'aqua add hpc2020 -e <path/to/your/catalogue>'"
```
mnurisso, line 144: `aqua add hpc2020 -e <path/to/your/catalogue>` will call the catalog hpc2020, maybe better `aqua add your-catalog-name -e <path/to/your/catalogue>`

mnurisso:
I left a last suggestion but we can merge this and work on a shared solution for new machines and where/how to document it in aqua

cli/hpc2020-install/hpc2020_install.sh
jhardenberg: done

mnurisso:
Merging, thanks!

PR #1786, opened 2025-03-03
Estimate time to generate weights only if regular grid

### LINKED ISSUE #1785: Making weights fails if target grid is not regular lat/lon
### machine

Lumi

### catalog

climatedt-phase1

### version

main

### What happened?

Currently all our grids used as target for regridding are ultimately using a regular lon/lat grid as a target (so for example r360x180, r180x90 etc in CDO format).
If we try to define new grids in `grids/default.yaml` which are not regular lat/lon (for example I added a line 'F128: F128' to allow using Gaussian grids, still in CDO format), the reader breaks when creating the weights. The reason is the code which computes an estimate of the length in time of the computation, which has a hardcoded assumption of regular grid in it.  The affected lines of code are the following in `regrid.py` which assume a 'rAAAxBB'  format for the target grid:

``` python
        width, height = map(int, cfg_regrid['grids'][regrid][1:].split('x'))
        new_grid_size = width * height
        
        total_size = sgrid.sizes
        total_elements = 1
        for dim_size in total_size.values():
            total_elements *= dim_size

        if original_grid_size > 0:  # Prevent division by zero
            vert_coord_size = total_elements / original_grid_size
        else:
            vert_coord_size = 1

        self._weights_generation_time(original_grid_size=original_grid_size,
                                      new_grid_size=new_grid_size, vert_coord_size=vert_coord_size, nproc=nproc)

```

I see four solutions:

1) drop this code entirely. Advantage: it works again and less code to maintain: Disadvantage: in some cases it might be good to get a warning that the calculation will be long and the estimate more or less works (at least in terms of ballpark estimate)

2) place this code in a separate user-callable function to estimate the weight calculation time and substitute it in `regrid.py` with a warning that the calculation may take some time in some cases and that they can use that function to make an estimate if desired. The function would then refuse to work if the grid is not rAAAxBBB. Advantage: we keep working code, we allow the code to work always. Disadvantage: long weight estimations may still come as a surprise to somebody. Hard to see that many people would actually use sucha function.

3) insert a check before this code so that it is run only if the target grid is in format rAAAxBBB and if not a simple warning log message like "Estimation of weight calculation time not possible because target is not a regular lon/lat grid. Please be aware that this may take a while." or similar. Advantage: we keep current functionality if possible

4) Try to make an estimate also for new grids different from regular. Disadvatages: .... many.

```python

```

### Error messages

_No response_

None

jhardenberg:
## PR description:

This addresses the issue in #1785 choosing option 3: the time to compute weights is calculated only if the target grid is regular lat lon.
A test would be nice but I don't know how to implement this because the ony effect if no time i estimated is to print. a debug log message. Do we have a way to check log messages in the tests? Any example?

Close #1785 

----

 - [ ] Changelog is updated.

### DIFF
--- a/src/aqua/reader/regrid.py
+++ b/src/aqua/reader/regrid.py
@@ -3,6 +3,7 @@
 import types
 import xarray as xr
 from smmregrid import CdoGenerate
+import re
 
 class RegridMixin():
     """Regridding mixin for the Reader class"""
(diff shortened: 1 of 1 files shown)


mnurisso:
@jhardenberg I think you can do similarly to `test_info.py`. The `capsys` from `pytest` will capture the output:

```python
@pytest.mark.aqua
def test_info_intake(capsys):
    """Test the info method of the Reader class with intake"""
    reader = Reader(model='NEMO', exp='test-eORCA1', source='long-2d',
                    regrid='r100', loglevel=loglevel)
    reader.info()
    captured = capsys.readouterr()

    assert "Reader for model NEMO, experiment test-eORCA1, source long-2d" in captured.out
```

I think for logger you need `captured.err`, here I was checking the info prints.

jhardenberg:
I added the test as discussed above and I profit from this to introduce a couple of additional target Gaussian target grids needed for the test (and which we will need anyway)

mnurisso:
I've updated test typo and documentation (we have a list of the target grids in the documentation for user qol). If tests are ok I'll merge (they are ok locally)

mnurisso:
I drop the `syscap` test for a more simple regrid with check on shape and values, since it's important to check the regrid to non lon lat and not only the logger.
For future reference, `syscap` doesn't work because in our `log_configure` we call `terminal = logging.StreamHandler()` without specifying anything, so that not `err` nor `out` get the output in the `syscap` returned dictionary

mnurisso:
Tests are now finally ok, merging

jhardenberg:
Thanks for fixing this and for merging!

PR #1790, opened 2025-03-03
Update AQUA attributes standard aspect

### LINKED ISSUE #1776: Introduction of AQUA specific attributes in xarray
### Feature request

In order to handle better attributes coming from aqua (model, exp, source) we propose to have a standard name for attribute, similar to what GRIB does.

model will become AQUA_model and so on.

We then need to update the implementation done in #1678 and add the AQUA_version
Add also source grid name if available

mnurisso:
## PR description:

Following the decision taken in #1776 I've updated the already existing attributes added by AQUA and added the version as required in #1772 

Close #1776 

 - [x] Changelog is updated.


oloapinivad (approved):
Looks good to me!

PR #1791, opened 2025-03-03
Remove usage of generator option 

### LINKED ISSUE #1770: Remove generator option from the Reader
We're going to remove the generator options from the Reader since it is unused.
A PR will be open and the removal evaluated to check if it is a consistent advantage or the code can be left.

silviacaprioli:
## PR description:

This PR remove the generator options from the Reader since it is unused.
The OPA_generator in the LRA folder is also removed

Close  #1770 

----

 - [x] Changelog is updated.


silviacaprioli:
Almost done here, but I have a couple of questions:

1. Not sure what to do with the `keep_memory` option in `fixer  `  
 `def fixer(self, data, destvar, apply_unit_fix=True, keep_memory=False)`
Is it only relevant for iterators, right? Should I remove it?

2. I haven't made any changes to intake_gsv.py; I only removed the usage of the iterator in Reader & co.
Is it fine? or should I revise also the intake_gsv?

@mnurisso @oloapinivad @jhardenberg

mnurisso:
> Not sure what to do with the keep_memory option in fixer  
def fixer(self, data, destvar, apply_unit_fix=True, keep_memory=False)
Is it only relevant for iterators, right? Should I remove it?

`keep_memory` according to docstring is also for decumulation, so it is still needed. Anyway in _wrapper_decumulate there is still a special case for iterators. Tagging @oloapinivad since he's the last one that worked on that part of code (line 709 in `fixer.py` from this branch)

> I haven't made any changes to intake_gsv.py; I only removed the usage of the iterator in Reader & co.
Is it fine? or should I revise also the intake_gsv?

From my inspection there is some if else with `self.dask_access`, but this value is always true. Tagging @jhardenberg for a second opinion. I think the strategy is if `to_dask` is called then `self.dask_access` exists otherwise no, but I don't think not having a value in the `__init__` is a clean way to use it.

Other places to check:

- `self.previous_data = None  # used for FDB iterator fixing` probably linked to the decumulation above, to be checked (initialization in `reader.py`)
- Mention of iterators in `_retrieve_plain `docstrings.

src/aqua/reader/reader.py
```diff
@@ -645,19 +629,13 @@ def set_default(self):
         Reader.instance = self  # Refresh the latest reader instance used
 
     def regrid(self, data):
```
mnurisso, line 631: This was a wrapper isn't it? There is need for the `regrid` and then the hidden `_regrid` methods?

mnurisso (commented):
Everything seems fine.
However in the end there is no modification in the intake_gsv, I'd be for cleaning also there, cc @jhardenberg

src/aqua/reader/reader.py
silviacaprioli: I agree, let's just rename the `_regrid` and remove the wrapper

silviacaprioli:
> Everything seems fine. However in the end there is no modification in the intake_gsv, I'd be for cleaning also there, cc @jhardenberg

me too, but since it's Jost creature I would wait for his green flag. Anyway he is on vacation until next monday...

silviacaprioli:
> > Everything seems fine. However in the end there is no modification in the intake_gsv, I'd be for cleaning also there, cc @jhardenberg
> 
> me too, but since it's Jost creature I would wait for his green flag. Anyway he is on vacation until next monday...

Actually on monday meeting he said that it's better not touch it. Maybe we can merge this PR without the intake_gsv cleaning and open a new issue specifically for that

jhardenberg:
Hmmm... is there actually any generator related code in intake-gsv?
Do you have an example?


Il Mar 11 Mar 2025, 20:56 Silvia Caprioli ***@***.***> ha
scritto:

silviacaprioli:
> Hmmm... is there actually any generator related code in intake-gsv? Do you have an example? 
[here ](https://github.com/DestinE-Climate-DT/AQUA/blob/476c78549f904800efbd87546a43f4e964799aea/src/aqua/gsv/intake_gsv.py#L644) for example

and there is also a `fstream_iterator` flag

jhardenberg:
`fstream_iterator` is a GSV interface thing defining a special mode to be used for bridge access. So, while in the future we may need to reconsider if it is still needed, it is not relevant for the issue here.

`read_chunked()` is a standard intake method providing an iterator. While it is true that we are not using it, I would keep it since I stll have not given up on the idea one day to make intake_gsv an independent intake driver and then this would be a method which it is useful to provide.

mnurisso:
@jhardenberg so is the _get_schema internally used by intake and it's why you have a dask true and false block?

However I'm also quite happy if in the future we can move the intake_gsv module outside, but at the same time to translate it to intake 2 will totally change the methods (to_dask is not a standard anymore for example afaik)

So if no intake_gsv modification is needed I'd do another review round and approve the PR, we can then merge it monday

jhardenberg:
yes, `_get_schema()` is called by `._load_metadata()` in `base.DataSource` of intake.
There is a `.read()`method which in theory loads data in memory, no dask. But I also admit that these parts are not used and not very much tested, so if we want to make this a separate driver it would require some work to investigate this.

mnurisso (approved):
I'd be for a merge, the decision on the intake_gsv has been taken, if I don't see any complain I'll soon merge it.

silviacaprioli:
> I'd be for a merge, the decision on the intake_gsv has been taken, if I don't see any complain I'll soon merge it.

Yes I think this can be merged

mnurisso:
Merging

PR #1800, opened 2025-03-05
Fix tcc grib code and add cmor sources to the convention file

### LINKED ISSUE #1798: Wrong grib code for tcc in convention-eccodes
### machine

Lumi

### catalog

climatedt-phase1

### version

main

### What happened?

The grib code for total cloud cover (tcc) in `convention-eccodes.yaml` is currently set to 131164. This, according to the ECMWF parameter database is actually tccp 'Total cloud cover probability` which is something else. the correct grib code, which we also have in the production protfolio, is 228164 (tcc). 

```python

```

### Error messages

_No response_

None

jhardenberg:
## PR description:

Add some cmor sources to `convention-eccodes.yaml` which were missing.
This also fixes a wrong grib code discussed in #1798 

Close #1798

----

 - [x] Changelog is updated.

### DIFF
--- a/config/fixes/convention-eccodes.yaml
+++ b/config/fixes/convention-eccodes.yaml
@@ -65,7 +65,6 @@ convention_name:
           - avg_hcc
           - avg_hcc_frac
           - HCC # AQUA ERA5
-          - clh # cmor
         grib: 3075
       lcc: # Low cloud cover
         source:
@@ -77,7 +76,6 @@ convention_name:
           - avg_lcc
           - avg_lcc_frac
           - LCC # AQUA ERA5
-          - cll # cmor
         grib: 3073
       mcc: # Medium cloud cover
         source:
@@ -89,7 +87,6 @@ convention_name:
           - avg_mcc
           - avg_mcc_frac
           - MCC # AQUA ERA5
-          - clm # cmor
         grib: 3074
       msl: # Mean sea level pressure
         source:
@@ -209,7 +206,6 @@ convention_name:
           - tciw
           - avg_tciw
           - clivi # vertically integrated cloud ice (https://easy.gems.dkrz.de/DYAMOND/NextGEMS/Models/c2_icon_ngc2013_output.html)
-          - iwp
         grib: 79
       tclw: # Total column vertically-integrated cloud liquid water
         source:
--- a/config/fixes/convention-eccodes.yaml
+++ b/config/fixes/convention-eccodes.yaml
@@ -13,38 +13,39 @@ convention_name:
           - 2d
           - avg_2d
           - D2M # AQUA ERA5
+          - tdps # cmor
         grib: 168
       2t: # 2 meter temperature
         source:
           - 167 # original GRIB2 code
           - 228004
           - 2t
-          - tas # cmor
           - mean2t # legacy eccodes
           - avg_2t
           - T2M # AQUA ERA5
+          - tas # cmor
         grib: 167
       10u: # 10 meter U wind component
         source:
           - 165 # original GRIB2 code
           - 235165 # avg_10u
           - 10u
-          - uas # cmor
           - m10u # legacy eccodes
           - avg_10u 
           - U10M # AQUA ERA5
           - u10 # cds
+          - uas # cmor
         grib: 165
       10v: # 10 meter V wind component
         source:
           - 166 # original GRIB2 code
           - 235166 # avg_10v
           - 10v
-          - vas # cmor
           - m10v # legacy eccodes
           - avg_10v
           - V10M # AQUA ERA5
           - v10 # cds
+          - vas # cmor
         grib: 166
       clwc: # Specific cloud liquid water content
         source:
@@ -52,6 +53,7 @@ convention_name:
           - 235246 # avg_clwc
           - clwc
           - avg_clwc
+          - lwp # cmor
         grib: 246
       hcc: # High cloud cover
         source:
@@ -63,6 +65,7 @@ convention_name:
           - avg_hcc
           - avg_hcc_frac
           - HCC # AQUA ERA5
+          - clh # cmor
         grib: 3075
       lcc: # Low cloud cover
         source:
@@ -74,6 +77,7 @@ convention_name:
           - avg_lcc
           - avg_lcc_frac
           - LCC # AQUA ERA5
+          - cll # cmor
         grib: 3073
       mcc: # Medium cloud cover
         source:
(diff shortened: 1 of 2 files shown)


config/fixes/convention-eccodes.yaml
```diff
@@ -74,6 +77,7 @@ convention_name:
           - avg_lcc
           - avg_lcc_frac
           - LCC # AQUA ERA5
+          - cll # cmor
```
oloapinivad, line 80: uhm, this does not look like a cmor standard https://github.com/search?q=repo%3APCMDI%2Fcmip6-cmor-tables%20cll&type=code

config/fixes/convention-eccodes.yaml
```diff
@@ -85,6 +89,7 @@ convention_name:
           - avg_mcc
           - avg_mcc_frac
           - MCC # AQUA ERA5
+          - clm # cmor
```
oloapinivad, line 92: as cll

config/fixes/convention-eccodes.yaml
```diff
@@ -179,28 +186,30 @@ convention_name:
...
+          - clt # cmor
+        grib: 228164
       tciw: # Total column vertically-integrated cloud ice water
         source:
           - 79 # original GRIB2 code
           - 235088 # avg corresponding to 79
           - tciw
           - avg_tciw
           - clivi # vertically integrated cloud ice (https://easy.gems.dkrz.de/DYAMOND/NextGEMS/Models/c2_icon_ngc2013_output.html)
+          - iwp
```
oloapinivad, line 212: I guess this is not cmor either, but where is coming from?

config/fixes/convention-eccodes.yaml
```diff
@@ -347,7 +358,7 @@ convention_name:
           - mslhf # legacy eccodes
           - slhf
           - SLHF # AQUA ERA5
-          - hfls # cmor
+          # - hfls # cmor --> sign wrong
```
oloapinivad, line 361: I am confused, this means we cannot use the convention here?

config/fixes/convention-eccodes.yaml
```diff
@@ -359,7 +370,6 @@ convention_name:
           - msnlwrf
           - str
           - STR # AQUA ERA5
```
oloapinivad, line 372: why you erased rlus?

config/fixes/convention-eccodes.yaml
```diff
@@ -416,6 +427,7 @@ convention_name:
           - mtnlwrf
           - ttr
           - TTR # AQUA ERA5
+          # - rlut # cmor --> sign wrong
```
oloapinivad, line 430: as above, what we do in this case?

mnurisso:
> I left some comments, I am mostly puzzled by the wrong sign case, what do we plan to do in that case?

In the long run I'd like to have not only the possibility to search for match as source block but also in the derived block to handle simple transformations like a sign change.

mnurisso:
So, from this PR the tcc fix should go in the operational, is it correct?

config/fixes/convention-eccodes.yaml
jhardenberg: Exactly. There is a range of fluxes which are positive in cmor (eg hfls)  and would have an opposite sign convention in grib. Currently all these are 'source's ... so they are just renamed. We have 'derived' but it does not accept lists for now.  Basically any variable requiring sign changes would needto be explicitly specified (as 'derived') in a fixe

config/fixes/convention-eccodes.yaml
jhardenberg: same reason as above: it would have a wrong sign!

config/fixes/convention-eccodes.yaml
jhardenberg: Either we live with it (which means that the convention does not cover these variables and a specific fix needs to be specified), or we allow also the 'derived' key to have a list of formulas (and they could then be included in the convention file, or we develop the possibility to specify a - sign also for 'sources' (basically making it a bit similar to 'derived')

jhardenberg:
> So, from this PR the tcc fix should go in the operational, is it correct?

Uh, yes, that you could be done also for OP, yes. It is really a wrong grib code being used

config/fixes/convention-eccodes.yaml
jhardenberg: You are right! This is a EC-Earth4 invention .... what should we do about these? Maybe NOT include them and leave them in their separate fixes file ?

jhardenberg: They are really not very standard and may change in future versions

config/fixes/convention-eccodes.yaml
jhardenberg: same as above, ece4

config/fixes/convention-eccodes.yaml
jhardenberg: again, ece4 invention. Sorry I misse dthese, it was actually my plan to keep these in a separate fixes file

jhardenberg: Ok, for now I removed them. If instead you think it is ok to have also EC-Earth4 'inventions' in here I will put them back

mnurisso:
I'm personally not so against having in the convention also some EC-EARTH4 specific varname if we're aware that it should not produce wrong results with other models.

About the derived quantities, it is something we definitely need to discuss, it can be done before or after the planning for a Fixer as a modular class, depending on the priority we feel for such a feature

jhardenberg:
Hi, to wrap up this one: should I put back EC-Earth specific 'inventions'? Thy are not entirely standard and I fear that they may change in the future as EC-Earth evolves ...

mnurisso:
> Hi, to wrap up this one: should I put back EC-Earth specific 'inventions'? Thy are not entirely standard and I fear that they may change in the future as EC-Earth evolves ...

If they may change then no, I thought they were an established EC-EARTH standard.

mnurisso:
I'd finalize this version to have asap tcc working in the main as well.
As discussed I'd leave out the EC-EARTH specific ones since they're possibly not definitive.
I'm more than glad to have them back here if they will become definitive and we know they do not create issues with the other models

jhardenberg:
@mnurisso I already had removed the ecearth-specific codes. I think you could merge this one as it is and we can add new ones in the future in a second iteration when needed.

mnurisso (approved):

mnurisso:
Ok, thanks, I'm merging this!

PR #1806, opened 2025-03-07
`convert_data()` as diagnostics core utility function

mnurisso:
## PR description:

I introduce a conver_data() function in the diagnostic utilities in order to convert data from one unit to another one, taking a dataset/dataarray and modifying only the requested variable. It also update the history attribute as the Reader does for data provenance.

 - [x] Changelog is updated.

### DIFF
--- a/src/aqua_diagnostics/core/util.py
+++ b/src/aqua_diagnostics/core/util.py
@@ -157,7 +157,7 @@ def merge_config_args(config: dict, args: argparse.Namespace,
     return config
 
 
-def convert_data(data, var: str, units: str, loglevel: str = 'WARNING'):
+def convert_data_units(data, var: str, units: str, loglevel: str = 'WARNING'):
     """
     Make sure that the data is in the correct units.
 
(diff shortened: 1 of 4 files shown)


oloapinivad:
Looks good but please not `convert_data` since is dealing with units, no?

mnurisso:
> Looks good but please not `convert_data` since is dealing with units, no?

ok, what do you suggest? It was check_data before, didn't sound good as well

src/aqua_diagnostics/core/util.py
```diff
@@ -154,3 +155,39 @@ def merge_config_args(config: dict, args: argparse.Namespace,
...
+    data_to_fix = data[var] if isinstance(data, xr.Dataset) else data
+    final_units = units
+    initial_units = data_to_fix.units
+
+    conversion = convert_units(initial_units, final_units)
+
+    factor = conversion.get('factor', 1)
+    offset = conversion.get('offset', 0)
+
+    if factor != 1 or offset != 0:
```
oloapinivad, line 180: I would return `data_to_fix` here if the condition is not met so you save a bit of time

oloapinivad:
> ok, what do you suggest? It was check_data before, didn't sound good as well

`check_data_units`? or `convert_data_units`? I mean everything that shows that we are using units

src/aqua_diagnostics/core/util.py
mnurisso: right

oloapinivad (approved):

mnurisso:
Totally forgot this was done and approved. Merging

PR #1814, opened 2025-03-10
[OP] Machines for tropical rainfall in the operational cycle

mnurisso:
## PR description:

Adding lumi-o25.1 and MN5-o25.1 for the tropical rainfall diagnostic

 - [x] Changelog is updated.


mnurisso:
I've tested on lumi with a lumi-e25.1 machine. It should work on the workflow as well so I'd go for a merge

oloapinivad (approved):

Issue #1788, opened 2025-03-03
Check usage of time handling in the current AQUA implementation


mnurisso:
Starting from #1779 we propose to first take a look to the pieces of code in which the time is handled and to comment if something that could be improved is present.

@mcadau will take a look and report in a PR with comments on that parts of code.

Issue #1779, opened 2025-02-28
Refactor Time Handling: Remove Deprecated datetime Objects and Standardize Usage


mnurisso:
### Feature request

After an offline discussion we propose here how to proceed with unification of times handling in AQUA

## Summary of Key Learnings

1. **Deprecation of `datetime` Objects**
   - `datetime` objects should be removed from the code.
2. **Interchangeability of `np.datetime64` and `pd.Timestamp`**
   - These two types are functionally equivalent.
3. **Distinction Between `pd.Timedelta` and `pd.DateOffset`**
   - `pd.Timedelta`: Used for specific time differences.
   - `pd.DateOffset`: Used for calendar-aware operations (no NumPy equivalent).
   - `DateOffset` only accepts arguments as keys, while `to_offset` can convert pandas frequency strings.
4. **DatetimeIndex as the Standard for Arrays**
   - `DatetimeIndex` is an array of `datetime64`, making arrays of `pd.Timestamp` unnecessary.
   - `DatetimeIndex` is preferred for defining time arrays in xarray.
5. **Generating DatetimeIndex**
   - Use `pd.date_range` or `xr.date_range` to create `DatetimeIndex` objects.
6. **Handling xarray Time Objects**
   - `xarray` time objects are NumPy arrays of `datetime64`.
   - Always convert them to pandas objects before performing operations:
     - `pd.to_datetime(data.time)`
     - `data.time.to_index()`
7. **Conversion Rules**
   - A single `to_datetime` object is always a `Timestamp`.
   - Multiple values return a `DatetimeIndex`.
   - `pd.Timestamp` and `DatetimeIndex` can be converted to string(s) using `.strftime()`.
   - Convert strings to timestamps using `pd.to_datetime()`.
8. **TimedeltaIndex for Time Differences**
   - `TimedeltaIndex` is the time difference counterpart of `DatetimeIndex`.

## Proposed Code Changes

- Remove all instances of `datetime` objects in favor of `np.datetime64` or `pd.Timestamp`.
- Replace arrays of `pd.Timestamp` with `DatetimeIndex`.
- Ensure `pd.to_datetime()` is used when working with xarray time objects.
- Use `pd.DateOffset` for calendar-aware operations instead of `pd.Timedelta`.
- Ensure time differences are handled using `TimedeltaIndex`.

## Next Steps

- Identify all usages of `datetime` objects in the codebase.
- Refactor time handling according to the best practices listed above.
[...]

PR #1827, opened 2025-03-14
Refactor Time Handling: Part 2

mcadau:
## PR description:
Following our internal discussion, we have started to implement the unification of the AQUA times management.
Following issues #1779 and #1788 , a comprehensive assessment of how AQUA manages time objects is in progress.

In this PR the aim is to replace all the lists of `pd.Timestamp` with `DatetimeIndex` elements.
lists of `pd.Timestamp` elements are currently in: `gsv/timeutil.py`, which will be modified accordingly.

#1779 and #1788 (along with PR #1820)


mnurisso:
Hi @mcadau I see many commits that seem to be from the timeseries refactor. Could you check if this is the case and eventually open a PR where only the time handling commits are present?

PR #1832, opened 2025-03-18
`TimStat` refactor

oloapinivad:
## PR description:

This creates a specific `TimStat` module independent from the `Reader()`. So far the modifications are minimals and most of the core methods are exposed through the Reader itself. 

 - [x] Changelog is updated.


src/aqua/reader/timstat.py
```diff
@@ -1,57 +1,36 @@
...
-                            time_bounds=time_bounds, center_time=center_time)
-
-    def timstd(self, data, freq=None, exclude_incomplete=False,
-               time_bounds=False, center_time=False):
-        """
-        Exposed method for time standard deviation statistic. Wrapper for timstat()
+    def timstat(self, data, stat='mean', freq=None, exclude_incomplete=False,
+                time_bounds=False, center_time=False):
+        
+        """"
```
mnurisso, line 24: I'd modify this into a proper docstring with args and returned values

mnurisso (approved):
It seems quite a natural implementation, we can go with a merge

mnurisso:
I'd modify the time statistic part of documentation mentioning the class availability: https://aqua.readthedocs.io/en/latest/other_components.html#time-statistics
Then we can double check cleaning and linting but this is ready to be merged

mnurisso:
> Thanks for the review, will do both things. Do you think it is worth it to move it to another folder out from the reader one?

ah, good point, yes, even if they are few lines of code but they are conceptually not Reader

src/aqua/reader/timstat.py
oloapinivad: Added the docstring

oloapinivad:
I made some updates and cleaning up of old code, which should be a bit more simple and following the idea of building entirely on pands #1779

oloapinivad:
If tests are passed, this can be merged for me

PR #1833, opened 2025-03-18
update base container to eccodes 2.40

oloapinivad:
## PR description:

This PR updates the docker base creation to eccodes 2.40. After container generation, workflow files need to be updated as well.


mnurisso:
Tests are done, merging

PR #1834, opened 2025-03-18
Check if GID exists before adding group in Dockerfile-development

mnurisso:
## PR description:

Since ubuntu 24.04 the GID is set to 1000 by default in the container. We then update the dockerfile to set GID only if this is not defined yet.

 - [x] Changelog is updated.

### DIFF
--- a/dockerfiles/Dockerfile-development
+++ b/dockerfiles/Dockerfile-development
@@ -22,7 +22,6 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
 # We switch to mamba user later, root for now just to create the user.
 USER root
 
-# --- x snip x ---
 # --- This is from the official docs from the micromamba image.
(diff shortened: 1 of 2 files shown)


dockerfiles/Dockerfile-development
```diff
@@ -42,7 +41,13 @@ COPY --from=micromamba /usr/local/bin/_entrypoint.sh /usr/local/bin/_entrypoint.
 COPY --from=micromamba /usr/local/bin/_dockerfile_initialize_user_accounts.sh /usr/local/bin/_dockerfile_initialize_user_accounts.sh
 COPY --from=micromamba /usr/local/bin/_dockerfile_setup_root_prefix.sh /usr/local/bin/_dockerfile_setup_root_prefix.sh
 
-RUN /usr/local/bin/_dockerfile_initialize_user_accounts.sh && \
+#FIX: Check if GID exists before adding group
```
oloapinivad, line 44: add that this is required by ubuntu 24.04 for future memory

mnurisso: donve

PR #1835, opened 2025-03-18
Introducing a `fldstat` module

oloapinivad:
## PR description:

This will move the fldmean function from the Reader creating a specific `FldStat` module that can be later extended for further horizontal dimension operations. 

There are many issues though, the most important ones are 
1. Wrong horizontal dimension naming between the areas (i.e. the source grid file) and the data
2. Wrong alignment of data models that cause latitudes and longitudes to have not the same orientation
3. Some structural confusion between dimensions and coordinates, where the first are used for spatial averaging and the second are used for area selection 

 - [ ] Changelog is updated.

### DIFF
--- a/src/aqua/fldstat/fldstat.py
+++ b/src/aqua/fldstat/fldstat.py
@@ -1,0 +1,102 @@
+"""AQUA class for field statitics"""
+import xarray as xr
+
+from aqua.logger import log_configure, log_history
+from aqua.util import area_selection
+
+class FldStat():
+    """AQUA class for field statitics"""
+
+    def __init__(self, area, horizontal_dims=None, grid_name=None, loglevel='WARNING'):
+        """
+        Initialize the FldStat.
+
+        Args:
+            area (str): The area to calculate the statistics for.
+            loglevel (str): The logging level.
+        """
+
+        self.loglevel = loglevel
+        self.logger = log_configure(log_level=loglevel, log_name='FldStat')
+        self.area = area
+
+        if self.area is None:
+            self.logger.warning("No area provided, no weighted area can be provided.")
+            return
+
+        # safety checks
+        if not isinstance(area, (xr.DataArray, xr.Dataset)):
+            raise ValueError("Area must be an xarray DataArray or Dataset.")
+        
+        # TODO: add a guess with smmregrid GridInspector
+        if horizontal_dims is None:
+            self.logger.warning("No horizontal coordinates provided, using default ['lon', 'lat'].")
+            horizontal_dims = ['lon', 'lat']
+        self.horizontal_dims = horizontal_dims
+        self.logger.debug('Space coordinates are %s', self.horizontal_dims)
+        self.grid_name = grid_name
+
+        
+    def fldmean(self, data, lon_limits=None, lat_limits=None, **kwargs):
+        """
+        Perform a weighted global average.
+        If a subset of the data is provided, the average is performed only on the subset.
+
+        Arguments:
+            data (xr.DataArray or xarray.DataDataset):  the input data
+            lon_limits (list, optional):  the longitude limits of the subset
+            lat_limits (list, optional):  the latitude limits of the subset
+
+        Kwargs:
+            - box_brd (bool,opt): choose if coordinates are comprised or not in area selection.
+                                  Default is True
+
+        Returns:
+            the value of the averaged field
+        """
+
+        if not isinstance(data, (xr.DataArray, xr.Dataset)):
+            raise ValueError("Data must be an xarray DataArray or Dataset.")
+
+        if self.area is None:
+            return data.mean(dim=self.horizontal_dims)
+
+        if lon_limits is not None or lat_limits is not None:
+            data = area_selection(data, lon=lon_limits, lat=lat_limits,
+                                  loglevel=self.loglevel, **kwargs)
+
+        # cleaning coordinates which have "multiple" coordinates in their own definition
+        # grid_area = self._clean_spourious_coords(grid_area, name = "area")
+        # data = self._clean_spourious_coords(data, name = "data")
+
+        # check if coordinates are aligned
+        try:
+            xr.align(self.area, data, join='exact')
+        except ValueError as err:
+            raise ValueError('Coordinates are not aligned!') from err
+            # # check in the dimensions what is wrong
+            # for coord in grid_area.coords:
+            #     if coord in space_coord:
+            #         xcoord = data.coords[coord]
+
+            #         # first case: shape different
+            #         if len(grid_area[coord]) != len(xcoord):
+            #             raise ValueError(f'{coord} has different shape between area files and your dataset.'
+            #                              'If using the LRA, try setting the regrid=r100 option') from err
+            #         # shape are ok, but coords are different
+            #         if not grid_area[coord].equals(xcoord):
+            #             # if they are fine when sorted, there is a sorting mismatch
+            #             if grid_area[coord].sortby(coord).equals(xcoord.sortby(coord)):
+            #                 self.logger.warning('%s is sorted in different way between area files and your dataset. Flipping it!',
+            #                                     coord)
+            #                 grid_area = grid_area.reindex({coord: list(reversed(grid_area[coord]))})
+            #             else:
+            #                 raise ValueError(f'{coord} has a mismatch in coordinate values!') from err
+
+        self.logger.debug('Computing the weighted average over f{self.horizontal_dims}')
+        out = data.weighted(weights=self.area.fillna(0)).mean(dim=self.horizontal_dims)
+
+        if self.grid_name is not None:
+            log_history(data, f"Spatially averaged by fldmean from {self.grid_name} grid")
+
+        return out
--- a/src/aqua/fldstat/fldstat.py
+++ b/src/aqua/fldstat/fldstat.py
@@ -30,7 +30,7 @@ class FldStat():
         
         # TODO: add a guess with smmregrid GridInspector
         if horizontal_dims is None:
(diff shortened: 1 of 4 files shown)


oloapinivad:
I somehow disagree. Point 2 is very likely associated to properties of the data model so it can be fixed "on line", but here the problem is slightly different. As far as I understand, data models conceptually should fix properties of the coordinates, here we are talking about the dimensions. 

I see two possible implementations here, but I am fully open to different suggestions: 
a. fixing the cell area when they are generated, as it was originally done. This can be done by loading the data, comparing the dimennsions length and naming, and adjust them accordingly. However, this requires the `Regridder()` to ALWAYS be initialized with data, so that it can compare data vs area dimensions. 
b. fixing the cell areas when they are load by the reader. This is an easier implementation, and more functional to the current structure, but would require a call to `_retrieve_plain()` for each new `Reader()`initialization, with a considerable slow down of the code.

oloapinivad:
Following a discussion offline with @mnurisso a possible solution to minimize the overhead is to create an attribute in the Regridder which provides information IF the areas have been generated with cdo (and not simply loaded from file) 
This attribute can then extracted by the `Reader` which can then make a call to `_retrieve_plain()` and then save the data again once the dimensions have been fixed. 

Only downside would be that there is no difference between fixed and non-fixed data. This was the behaviour closer to the original one.

pyproject.toml
```diff
@@ -107,7 +107,7 @@ version = {attr = "aqua.version.__version__"}
 
 [tool.pytest.ini_options]
 # verbose to have full test, run coverage on src folder and use pyproject as a configuration file
-addopts = "-v --cov=aqua"
+addopts = "--durations=10 -v --cov=aqua"
```
mnurisso, line 110: what is it?

src/aqua/fldstat/fldstat.py
```diff
@@ -0,0 +1,156 @@
...
+from smmregrid import GridInspector
+
+from aqua.logger import log_configure, log_history
+from aqua.util import area_selection
+
+
+class FldStat():
+    """AQUA class for field statitics"""
+
+    def __init__(self, area=None, horizontal_dims=None, grid_name=None, loglevel='WARNING'):
```
mnurisso, line 14: In general I'm starting to use the type requirement in the function/class arguments as well when clear.
Also by having None on all of them is not clear if any of them is mandatory. The init suggest that are all optional, while the docstring that only the grid_name is optional.

src/aqua/fldstat/fldstat.py
```diff
@@ -0,0 +1,156 @@
...
+        """
+
+        self.loglevel = loglevel
+        self.logger = log_configure(log_level=loglevel, log_name='FldStat')
+        self.area = area
+        if horizontal_dims is None:
+            self.logger.warning("No horizontal dimensions provided, will try to guess from data when provided!")
+        self.horizontal_dims = horizontal_dims
+
+        if self.area is None:
```
mnurisso, line 32: By reading this, I understand that the area is mandatory. I can understand the necessity of not having custom raise, but if you leave the init with area mandatory the error message will be clear enough, I'd personally go for this option

src/aqua/fldstat/fldstat.py
```diff
@@ -0,0 +1,156 @@
...
+
+class FldStat():
+    """AQUA class for field statitics"""
+
+    def __init__(self, area=None, horizontal_dims=None, grid_name=None, loglevel='WARNING'):
+        """
+        Initialize the FldStat.
+
+        Args:
+            area (str): The area to calculate the statistics for.
```
mnurisso, line 19: This is wrong, correct? Below it is written that it should be a DataArray or DataSet

src/aqua/fldstat/fldstat.py
```diff
@@ -0,0 +1,156 @@
...
+
+        if self.area is None:
+            self.logger.warning("No area provided, no weighted area can be provided.")
+            return
+
+        # safety checks
+        if not isinstance(area, (xr.DataArray, xr.Dataset)):
+            raise ValueError("Area must be an xarray DataArray or Dataset.")
+
+        self.logger.debug('Horizontal dimensions are %s', self.horizontal_dims)
```
mnurisso, line 40: Is this for your debug? There has been no guess, so the variable should be the same as the one provided in the init, am I wrong?

src/aqua/fldstat/fldstat.py
```diff
@@ -0,0 +1,156 @@
...
+
+        if not isinstance(data, (xr.DataArray, xr.Dataset)):
+            raise ValueError("Data must be an xarray DataArray or Dataset.")
+
+        # if horizontal_dims is not provided, try to guess it
+        if self.horizontal_dims is None:
+            data_gridtype = GridInspector(data).get_grid_info()
+            if len(data_gridtype) > 1:
+                raise ValueError("Multiple grid types found in the data, please provide horizontal_dims!")
+            self.horizontal_dims = data_gridtype[0].horizontal_dims
```
mnurisso, line 69: I'd add for future reference a comment that the data_gridtype is always a list by GridInspector construction, otherwise this step is not very clear

src/aqua/fldstat/fldstat.py
```diff
@@ -0,0 +1,156 @@
...
+            self.logger.debug('Horizontal dimensions guessed from data are %s', self.horizontal_dims)
+
+        #if area is not provided, return the raw mean
+        if self.area is None:
```
mnurisso, line 73: no logger of any form? This is unweighted mean, am I right? So a logger.warning should be necessary.
Do we want to return this or raise errors?

src/aqua/fldstat/fldstat.py
```diff
@@ -0,0 +1,156 @@
...
+            return data.mean(dim=self.horizontal_dims)
+        
+        # align dimensions naming of area to match data
+        self.area = self.align_area_dimensions(data)
+
+        # align coordinates values of area to match data
+        self.area = self.align_area_coordinates(data)
+
+        if lon_limits is not None or lat_limits is not None:
+            data = area_selection(data, lon=lon_limits, lat=lat_limits,
```
mnurisso, line 83: I leave it here as a comment, but the area_selection is not quite used by diagnostics. It would be nice to understand if this can be part of some module more than a random function to import if you know that exist.

src/aqua/reader/reader.py
```diff
@@ -181,6 +186,19 @@ def __init__(self, model=None, exp=None, source=None, catalog=None,
         # init the regridder and the areas
         self._configure_regridder(machine_paths, regrid=regrid, areas=areas,
                                   rebuild=rebuild, reader_kwargs=reader_kwargs)
+        
+        # init the fldstat modules. if areas are not available, will issue a warning
+        cell_area = self.src_grid_area.cell_area if areas else None
+        self.src_fldstat = FldStat(
```
mnurisso, line 192: Is it still there the option to set areas=False at the Reader initialization? I don't see any if here, the FldStat class is initialized anyway.

tests/test_fldmean.py
```diff
@@ -1,31 +1,59 @@
...
+class TestFldModule():
+    """Class for fldmean standalone"""
+
+    def test_fldmean_from_data(self):
+        """test Fldmean class native from data"""
+
+        reader = Reader(catalog='ci', model='IFS', exp='test-tco79',
+                        source='short', regrid='r100', rebuild=True)
+        data = reader.retrieve()
+        fldmodule = FldStat(area=reader.src_grid_area.cell_area, loglevel=LOGLEVEL)
```
mnurisso, line 19: why to use the FldStat externally and not the one initialized inside the Reader?

mnurisso (commented):
I left comments, a bit more than in the TimStat case. I think we can go for a merge very soon, in order to have all the new pieces in place and wait for bugs.
However I'd go then very quickly for adding other statistics to this module, specular to what is TimStat now.

pyproject.toml
oloapinivad: a cool feature that will report the 10 slowest test. it can be used for benchmarking on a local machine when doing development.

src/aqua/fldstat/fldstat.py
oloapinivad: acutally they are all optional. the module is meant to work also when no area is provided by doing a non-weighted average. I am adding the type requirements

src/aqua/fldstat/fldstat.py
oloapinivad: I might have misunderstood your comment, but no, the module will work anyway by doing spatial averages (and possibly other operations). it will simply not use the areas if they are not available but log a warning.

src/aqua/fldstat/fldstat.py
oloapinivad: yes I am fixing it

src/aqua/fldstat/fldstat.py
oloapinivad: yes I can remove it

src/aqua/fldstat/fldstat.py
oloapinivad: I agree it can be integrated in the fldstat maybe, or maybe having a class on its own

oloapinivad: totally agree

src/aqua/fldstat/fldstat.py
```diff
@@ -0,0 +1,181 @@
...
+
+from aqua.logger import log_configure, log_history
+from aqua.util import area_selection
+
+
+class FldStat():
+    """AQUA class for field statitics"""
+
+    def __init__(self,
+             area: xr.Dataset | xr.DataArray | None = None,
```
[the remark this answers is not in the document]
mnurisso: I just realized we already dropped python 3.9, this can be dropped as requested change

Issue #1685, opened 2025-02-08
OutputSaver assumes lra-r100-monthly source if no catalog specified


jhardenberg:
### machine

None

### catalog

None

### version

main

### What happened?

The `OutputSaver` code contains a hack so that if no catalog is specified, it searches for it assuming that the source is `lra-r100-monthly` . This has the consequence that trying to plot something as simple as a timeseries of ERA5 data fails.
Any experiment source which has not `lra-r100-monthly` will fail.
The problem is in these lines of the `OutputSaver` code:
```
if catalog:
            self.catalog = catalog
        else:
            # HACK: This is a temporary solution to get the catalog with the complete triplet.
            # At the moment we assume that the source is the 'lra-r100-monthly'.
            config = ConfigPath()
            _ = config.deliver_intake_catalog(model=model, exp=exp, source='lra-r100-monthly')
            self.catalog = config.catalog
```

The example below produces the figure correctly but then crashes while trying to save the figure.

### Code to reproduce the issue

```python
from aqua.diagnostics import Timeseries
ts = Timeseries(var='2t', models=['ERA5'], exps=['era5'], sources=['monthly'], loglevel='debug')
ts.run()
```

### Are you interested in making a pull request?

None

mnurisso:
Closed by #1837

PR #1837, opened 2025-03-19
OutputSaver refactor

### LINKED ISSUE #1764: Cleanup of `OutputSaver`
### Feature request

As discussed offline, there is the need of cleaning some of the option of the `OutputSaver`.

- [ ] Lighter `__init__`, moving some required argument as argument in following method
- [ ] Update `diagnostic_product` as an argument and not as class attribute
- [ ] Add multimodel case option for figures
- [ ] Area to be named as `region`
- [ ] `_2` argument named as `_ref`
- [ ] review the filename keys, currently hard-coded.

silviacaprioli:
## PR description:

Outputsaver cleanup.
I have created a new class `OutputSaver_NEW`, as requested. Eventually, it will replace the old one.

- [x] Lighter __init__, moving some required argument as argument in following method
- [x] Update diagnostic_product as an argument and not as class attribute
- [x] Add multimodel case option for figures
- [x] Area to be named as region
- [x] _2 argument named as _ref
- [x] review the filename keys, currently hard-coded.

Close  #1764  #1685 

----

 - [ ] Changelog is updated.


silviacaprioli:
skeleton of the class for the moment (many things missing), but I'm trying to keep it as minimalistic as possible

mnurisso:
A couple of comments:

1. we should decide where the tool goes. Should it stay in the framework or in the `aqua_diagnostics`?
2. I think we want to remain with the dots splitting the keys, so no underscore between then since it's quite used internally in keys
3. what if I have more models in my png or pdf file? At the moment I have to initialize the entire class with a model, but this may not be the case for png or pdf, where I may have multimodel plot. What's the strategy for this?

mnurisso:
> > 1. we should decide where the tool goes. Should it stay in the framework or in the `aqua_diagnostics`?
> 
> Since it is specific for saving diagnostic products, maybe putting it in aqua_diagnostic is a better option indeed...

This was my initial idea, so that you can call it OutputSaver in the new place and we switch the import one diagnostic at a time.

> > 3. what if I have more models in my png or pdf file? At the moment I have to initialize the entire class with a model, but this may not be the case for png or pdf, where I may have multimodel plot. What's the strategy for this?
> 
> Not sure about that, options could be:
> 
>     1. Modify the `generate_name` method to allow overriding `self.model` by passing it as an argument in `extra_keys`:
>        `extra_keys={'model': 'multi'}` or explicitly list the models.
> 
>     2. Instead of treating model as a single string, allowing it to be a list

The issue is that I use the OutputSaver in one situation: when I save the netcdf, where I need to store a single model and the model/exp are clearly useful. Then for the plot, where I can have more models and these args may have not so much meaning.

silviacaprioli:
> The issue is that I use the OutputSaver in one situation: when I save the netcdf, where I need to store a single model and the model/exp are clearly useful. Then for the plot, where I can have more models and these args may have not so much meaning.

ok we can modify the `generate_name` method so that:

1) If a single model is specified, exp and catalog are included as usual.
2) If `extra_keys={"model": "multi"}` or multiple models are specified (as a list, like ["IFS-NEMO", "ICON", "ERA5"]), exp and catalog are excluded to avoid ambiguity and the final name would be something like:

`dummy.mean.IFS-NEMO_ICON_ERA5.tprate
`or 
`dummy.mean.multi.tprate
`

I'll try to implement it and then we can see if it's a clean way to deal with it

silviacaprioli:
So, with the current implementation you can do the following:

```
output_saver = OutputSaver(diagnostic='GlobalBiases', model='IFS-NEMO', exp='historical', loglevel='DEBUG')
extra_keys= {'var': '2t'}
output_saver.save_netcdf(bias, diagnostic_product = 'totalbias', extra_keys=extra_keys)
```

'./netcdf/GlobalBiases.totalbias.IFS-NEMO.historical.2t.nc'

while if you need to save a pdf with multiple models you can change the extra_keys:

```
extra_keys = {'var': '2t', 'model' : ['IFS-NEMO', 'ERA5']}
output_saver.save_pdf(fig, diagnostic_product = 'totalbias', extra_keys=extra_keys)

```

'./pdf/GlobalBiases.totalbias.IFS-NEMO_ERA5.2t.pdf'


@mnurisso let me know if you think it can work

mnurisso:
It seems nice to me, way cleaner!
We can go for this strategy and review it next in person meeting, what do you think?

silviacaprioli:
Great! there are few things still to be added (like metadata) but I'll keep this general structure then

silviacaprioli:
I introduced the 'multimodel' and 'multiref' case, for example:
```

output_saver = OutputSaver(diagnostic='dummy', 
                           model=['IFS-NEMO','ICON'] , exp=['historical', 'historical'], catalog=['lumi-phase2', 'lumi-phase2'],
                           model_ref = 'ERA5', exp_ref = 'era5', catalog_ref = 'obs', loglevel='DEBUG')

extra_keys = {
        'region': 'indian_ocean'
    }

filename = output_saver.generate_name(diagnostic_product='mean', extra_keys=extra_keys)
```

gives: dummy.mean.multimodel.obs.ERA5.era5.indian_ocean

silviacaprioli:
I think this is now ready to be tested within the diagnostics.
The documentation hasn't been updated yet, I would  wait until the new design approach is confirmed and the old OutputSaver class is deprecated.

silviacaprioli:
merging, as agreed

Issue #1789, opened 2025-03-03
Intake v2 migration issues investigation


mnurisso:
We will use this issue to report on the possible issues in the update of intake package to v2.
In particular we need to check:

- [ ] integration of intake-xarray
- [ ] integration of intake-esm
- [ ] step needed for intake-gsv

mnurisso:
The call:

`reader = Reader(model='NEMO', exp='test-eORCA1', source='long-2d')`

produces:

```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[6], line 1
----> 1 reader = Reader('NEMO','test-eORCA1','long-2d')

File ~/AQUA/src/aqua/reader/reader.py:165, in Reader.__init__(self, model, exp, source, catalog, fix, regrid, regrid_method, areas, datamodel, streaming, stream_generator, startdate, enddate, rebuild, loglevel, nproc, aggregation, chunks, preproc, convention, **kwargs)
    163 # Manual safety check for netcdf sources (see #943), we output a more meaningful error message
    164 if isinstance(self.esmcat, intake_xarray.netcdf.NetCDFSource):
--> 165     if not files_exist(self.esmcat.urlpath):
    166         raise NoDataError(f"No NetCDF files available for {self.model} {self.exp} {self.source}, please check the urlpath: {self.esmcat.urlpath}") # noqa E501
    168 # store the kwargs for further usage

AttributeError: 'NetCDFSource' object has no attribute 'urlpath'
```

mnurisso:
More deeply, if I open the catalog with pure intake:

`cat = intake.from_yaml_file('/Users/mnurisso/.aqua/catalogs/ci/catalog.yaml')`

I get:

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[16], line 1
----> 1 cat = intake.from_yaml_file('/Users/mnurisso/.aqua/catalogs/ci/catalog.yaml')

File /opt/homebrew/Caskroom/mambaforge/base/envs/aqua-intake2/lib/python3.12/site-packages/intake/readers/entry.py:433, in Catalog.from_yaml_file(path, **kwargs)
    431 of = fsspec.open(path, **storage_options)
    432 with of as stream:
--> 433     cat = Catalog.from_dict(yaml.safe_load(stream))
    434 cat.user_parameters["CATALOG_PATH"] = path
[...]

PR #1838, opened 2025-03-20
Support for python 3.13

oloapinivad:
## PR description:

As per title, tests offline seems to work checkin on CI/CD

 - [ ] Changelog is updated.


oloapinivad:
So the environment creation works well, but unfortunately we need to upgrade `intake_esm` and very likely move to `intake2`. However, as discussed in #1789, this is not easy.

mnurisso:
I think this PR can be closed, the needed discussion has to happen in #1789

PR #1854, opened 2025-03-31
Extend map function to other projections

### LINKED ISSUE #1720: Diagnostics Graphics Refactoring: 2d maps
The aim is to build a 2d plot class for diagnostics which present 2d maps as output plots.

mnurisso:
## PR description:

We're expanding plot_single_map to other projections.

Close #1720 

----

 - [x] Documentation is included if a new feature is included. -> pictures to update
 - [x] Changelog is updated.


jhardenberg (approved):
looks great

src/aqua/util/graphics.py
```diff
@@ -176,19 +174,46 @@ def set_map_title(data: xr.DataArray, title: str = None,
...
+
+        try:
+            units = getattr(data, 'units', None)
         except AttributeError:
-            if model is not None and exp is not None:
-                title = f"{model} {exp}"
-                logger.debug("Using model and exp as map title")
+            units = None
+
+        try:
```
oloapinivad, line 191: what about if `model=data.attrs["AQUA_model'] if 'AQUA_model' in data.attrs else None`

src/aqua/util/graphics.py
```diff
@@ -176,19 +174,46 @@ def set_map_title(data: xr.DataArray, title: str = None,
...
-        try:
-            title = data.long_name
-            logger.debug("Using long_name as map title")
-
+        title = ""
+        for attr in ['long_name', 'short_name', 'shortName']:
             try:
-                logger.debug("Adding time to map title")
-                title = f"{title} {data.time.values}"
+                varname = getattr(data, attr, None)
```
oloapinivad, line 180: same here, is the try expect really necessary?

tests/test_graphics.py
```diff
@@ -17,74 +17,81 @@ def setup_method(self):
...
+                                  ticks_rounding=1,
+                                  cbar_ticks_rounding=1,
                                   loglevel=loglevel)
         assert fig is not None
         assert ax is not None
 
-        # Check the file was created
-        assert os.path.exists('tests/figures/test_single_map.png')
+        fig.savefig(tmp_path / 'test_plot_single_map.png')
+        assert os.path.exists(tmp_path / 'test_plot_single_map.png')
```
oloapinivad, line 49: is this a `Path` object? how can you join them in this way? curiosity

src/aqua/util/graphics.py
mnurisso: Sure make sense I'll add it in #1865

tests/test_graphics.py
mnurisso: They are `pathlib.Path` objects and this way to concatenate them is the suggested one in pytest docs so I just stick with it.

oloapinivad: Thanks,clear! We would need to decide sooner or later if we want to use pathlib everywhere!

mnurisso: For the test I think this solution is essential because I noticed some test was successful locally because I didn't fully clean previous run plots, I think same for the diagnostics

PR #1859, opened 2025-04-01
Regrid method capabilities back in the Reader

mnurisso:
## PR description:

As noticed offline, OSI-SAF data do not regrid because 'ycon' method is used by default. The regrid_method capabilities were lost in the recent Regridder update and are then reintroduced here.

 - [ ] Test 
 - [ ] Changelog is updated.


tovaz92 (approved):
Approved as agreed

PR #1862, opened 2025-04-01
AQUA internal data model for coordinate fixing

### LINKED ISSUE #1330: Data model refactoring
### Feature request

We are currently using the "CF2CDM" tool developed for cfgrib to enforce a 'data model' where coordinate units, direction and metadata are standardized. This approach is not entirely flexible since it recognizes automatically the coordinates, so that  the fixer is starting to have some overlapping functionality (see #1320 and #822).
We should refactor/rewrite the datamodel to be more flexible or decide which functionality leave in which tool.

oloapinivad:
## PR description:

This PR describes a tentative data model implementation which should tries to address most evident issues with coordinates, specifically addressing issues in longitude, latitude and - if possible - vertical coordinates too.

The idea is to work with two classes, one aiming at detecting the coordinates names and properties, and the other one working to convert toward a target convention to be used inside AQUA

Close #1330 

----

 - [x] Changelog is updated.


oloapinivad:
The new data model is in place, and should work as expected at least in most cases. However, there are many aspects which are not yet addressed but I would leave for future PRs (also because some of them touches the fixer) which has to be refactored similarly:
this can become a specific long term issue. 

- No data provenance is currently implemented in the data model. This should follow the usual AQUA standards
- We need to decide what to do when multiple coordinates are found along the same axis. We cannot operate the data model when two vertical coordinates, or even worse, two horizontal coordinates are found. What we want in that case? Disable the data model for that specific coordinate seems the safest option. 
- The data model has to be more modular. We probably need to introduce a type or a better a class object which defines the data model structure, a `CoordsObject()` which is nothing more that a structured dictionary. So that we have `CoordsIdentifier` which delivers the `CoordsObject()` to be used. Then `CoordsTransformer()` receives two object and the data to produce the metadata changes.
- Time coordinate detection by `CoordsIdentifier()` is not yet developed. Again, we need to decide what to do when multiple coordinates are found along time
- We should extend coordinate detection to model levels, height levels, and soil levels. 
- I would like to have more YAML based options to be read from files when working for coordinate detection rather than specific functions as done in `CoordsIdentifier()` so far. 
- We need to establish if we want the DataModel to run before the Fixer(), or after, and if we want the changes to be done at the end of the data loading or in the meantime. This concerns also the areas. 
- The depth dimension in FDB is not recognized by the datamodel, because having `m` as units it is too generic (this can be height as well) and there are no other useful info in the attributes. We need to ask this to the 
- We need to verify what happens when we modify a `dimension-coordinate` https://docs.xarray.dev/en/latest/user-guide/terminology.html#term-Dimension-coordinate. This might have impact in the Regridder() capabilities
- I have noticed extra units for the metpy are not read from file, but defined in a hard-coded function. This has to change, since it is not flexible.

oloapinivad:
@jhardenberg @mnurisso this waits for a review/approval before being merged

mnurisso:
> We need to decide what to do when multiple coordinates are found along the same axis

To notify the user and disable the data model seems the safest to me as well. If we will have real examples then we can decide if it's worth to have a more complex implementation

> The data model has to be more modular.

This is nice but requires in my opinion more knowledge on where the actual implementation may fail. I'd let the code to be tested a bit with real examples before

> Time coordinate detection by CoordsIdentifier() is not yet developed

This seems more urgent, since we have a real example on EC-EARTH4

> We should extend coordinate detection to model levels, height levels, and soil levels.

If we have real examples (I think we have some old source) this should have higher priority in my opinion

> I  would like to have more YAML based options to be read from files

I agree, it should be a simple task so I'd be for a high priority on this

> We need to establish if we want the DataModel to run before the Fixer()

I didn't check yet, how is it done now? I'd say at first glance data model first and then fixer, but I don't particularly know pro and cons of the two options

> The depth dimension in FDB is not recognized by the datamodel, because having m as units it is too generic (this can be height as well) and there are no other useful info in the attributes.

In this case the reader knows the stream in which data are written, do you think this can be exploited? Maybe the data model could take an argument that specify the realm if this is uniquely detected by the reader, so that any vertical coordinate is then clearly detected. I agree anyway that something added by the GSV would be a good solution and more clean.

src/aqua/data_model/coordidentifier.py
```diff
@@ -0,0 +1,189 @@
...
+    return False
+
+class CoordIdentifier():
+    """
+    Class to identify the nature of coordinates of an Xarray object.
+    It aims at detecting the longitude, latitude, time and any other vertical
+    by inspecting the attributes of the coordinates provided by the user.
+    """
+
+    def __init__(self, coords, loglevel='WARNING'):
```
mnurisso, line 35: you can add `coords: xr.Coordinates` to make clear to the user the required object

src/aqua/data_model/coordidentifier.py
```diff
@@ -0,0 +1,189 @@
...
+        """
+        Constructor of the CoordIdentifier class.
```
mnurisso, line 37: Docstring incomplete

src/aqua/data_model/coordidentifier.py
```diff
@@ -0,0 +1,189 @@
...
+        """
+        self.loglevel = loglevel
+        self.logger = log_configure(self.loglevel, 'CoordIdentifier')
+
+        if not isinstance(coords, xr.Coordinates):
+            raise TypeError("coords must be an Xarray Coordinates object.")
```
mnurisso, line 43: This could be easily tested, worth to add for coverage since it is very quick

src/aqua/data_model/coordidentifier.py
```diff
@@ -0,0 +1,189 @@
...
+            "depth": []
+        }
+
+    def identify_coords(self):
+        """
+        Identify the coordinates of the Xarray object.
+        """
+        for name, coord in self.coords.items():
+            self.logger.debug("Identifying coordinate: %s", name)
+            if self._identify_latitude(coord):
```
mnurisso, line 61: I know it shouldn't but this chain of if can potentially lead to a single coordinate identified as multiple of these categories, isn't it?

src/aqua/data_model/coordidentifier.py
```diff
@@ -0,0 +1,189 @@
...
+                self.coord_dict[key] = value[0]
+            elif len(value) > 1:
+                self.logger.error("Multiple %s coordinates found: %s. Using the first one: %s",
+                                     key, [x['name'] for x in value], value[0]['name'])
+                self.coord_dict[key] = value[0]
+        return self.coord_dict
+    
+    def _get_horizontal_attributes(self, coord):
+        """
+        Get the attributes of the coordinate.
```
mnurisso, line 96: Docstring

src/aqua/data_model/coordtranformer.py
```diff
@@ -0,0 +1,269 @@
...
+import xarray as xr
+from metpy.units import units
+from aqua.logger import log_configure, log_history
+from aqua.util import load_yaml
+from aqua import __path__ as pypath
+from .coordidentifier import CoordIdentifier
+
+
+# Function to get the conversion factor
+def units_conversion_factor(from_unit_str, to_unit_str):
```
mnurisso, line 13: Isn't it something that is potentially used by the fixer as well?

src/aqua/data_model/coordtranformer.py
```diff
@@ -0,0 +1,269 @@
...
+    """
+    Get the conversion factor between two units.
+    """
+    from_unit = units(from_unit_str)
+    to_unit = units(to_unit_str)
+    return from_unit.to(to_unit).magnitude
+
+# default target coords
```
mnurisso, line 21: Why hardcoded here?

src/aqua/data_model/coordtranformer.py
```diff
@@ -0,0 +1,269 @@
...
+    def __init__(self, data, loglevel='WARNING'):
+        """
+        Constructor of the CoordTransator class.
+
+        Args:
+            data (xr.Dataset or xr.DataArray): Xarray Dataset or DataArray object.
+            loglevel (str, optional): Log level. Defaults to 'WARNING'.
+        """
+        if not isinstance(data, (xr.Dataset, xr.DataArray)):
+            raise TypeError("data must be an Xarray Dataset or DataArray object.")
```
mnurisso, line 43: easy to test here as well

src/aqua/data_model/coordtranformer.py
```diff
@@ -0,0 +1,269 @@
...
+        self.loglevel = loglevel
+        self.logger = log_configure(self.loglevel, 'CoordTransator')
+
+        self.data = data
+
+        self.src_coords = CoordIdentifier(data.coords).identify_coords()
```
mnurisso, line 49: loglevel to give as argument

src/aqua/data_model/coordtranformer.py
```diff
@@ -0,0 +1,269 @@
...
+            if src_coord['bounds'] in data:
+                self.logger.info("Renaming bounds %s to %s",
+                                src_coord['bounds'], tgt_coord['bounds'])
+                data = data.rename({src_coord['bounds']: tgt_coord['bounds']})
+                data[tgt_coord['name']].attrs['bounds'] = tgt_coord['bounds']
+            else:
+                self.logger.info("Bounds %s not found in data.", src_coord['bounds'])
+        return data
+
+    def reverse_coordinate(self, data, src_coord, tgt_coord):
```
mnurisso, line 175: This seems to be poorly tested

src/aqua/data_model/coordtranformer.py
```diff
@@ -0,0 +1,269 @@
...
+                self.logger.info("Cannot reverse coordinate %s. Grid type is %s.",
+                                    tgt_coord['name'], self.gridtype)
+        return data
+    
+    
+    def convert_units(self, data, src_coord, tgt_coord):
+        """
+        Convert units of the coordinate.
+        """
+        if 'units' not in tgt_coord:
```
mnurisso, line 212: Do we have a specific case in mind where the target are supposed to have no units?

src/aqua/reader/fixer.py
```diff
@@ -1,20 +1,16 @@
...
 import pandas as pd
-from metpy.units import units
 
-from aqua.util import eval_formula, get_eccodes_attr, find_lat_dir
-from aqua.util import check_direction, to_list, normalize_units, convert_units
+from aqua.util import eval_formula, get_eccodes_attr
+from aqua.util import to_list, normalize_units, convert_units
 from aqua.logger import log_history
-from aqua.data_models import translate_coords
+#from aqua.data_models import translate_coords
```
mnurisso, line 12: remove this commented line

src/aqua/reader/fixer.py
```diff
@@ -601,9 +595,10 @@ def fixer(self, data, destvar, apply_unit_fix=True):
         # Fix coordinates according to a given data model
         src_datamodel = self.fixes.get("data_model", src_datamodel)
         if src_datamodel:
-            data = self.change_coord_datamodel(data, src_datamodel, self.dst_datamodel)
-            self.logger.info(f"coordinates adjusted to {src_datamodel} by AQUA fixer")
-            data = log_history(data, f"Coordinates adjusted to {src_datamodel} by fixer")
+            data = CoordTransformer(data, loglevel=self.loglevel).transform_coords()
+        #    data = self.change_coord_datamodel(data, src_datamodel, self.dst_datamodel)
```
mnurisso, line 599: All of them can be removed I suppose

src/aqua/util/coord.py
mnurisso, line 1: Is this file basically gone?

src/aqua_diagnostics/global_biases/global_biases.py
```diff
@@ -271,22 +271,25 @@ def plot_vertical_bias(self, data=None, data_ref=None, var_name=None, plev_min=N
         if plev_max is None:
             plev_max = bias['plev'].max().item()
 
-        bias = bias.sel(plev=slice(plev_max, plev_min))
+        if plev_min < plev_max:
```
mnurisso, line 274: It would be nice to compact this in one if else line to recover the correct coverage

mnurisso (approved):
I added a lot of small comments but the general structure seems totally fine to me.
I'd stress more in the documentation that a data model arg in the reader is present (I remember, if not maybe I'd add it) as True by default but this can be always disabled.

src/aqua/data_model/coordidentifier.py
oloapinivad: Done

src/aqua/data_model/coordidentifier.py
oloapinivad: I added a continue after each if so that this is not possible. it should also speed up the code

src/aqua/data_model/coordtranformer.py
oloapinivad: in principle yes. this is actually something that the fixer is using, but I wanted a self consistent tool

src/aqua/data_model/coordtranformer.py
oloapinivad: not sure I got the question here: we need to hardcode the name of the datamodel for now. It does not prevent you to use other data models though...

src/aqua/data_model/coordtranformer.py
oloapinivad: done

src/aqua/data_model/coordtranformer.py
oloapinivad: done

src/aqua/data_model/coordtranformer.py
oloapinivad: there is no validation of the tgt units since there is no class object for it. So in principle it is possible that a badly designed target has not units

src/aqua/reader/fixer.py
oloapinivad: done

src/aqua/util/coord.py
oloapinivad: yep, I can remove it

src/aqua_diagnostics/global_biases/global_biases.py
oloapinivad: this will is going to be modified by Silvia in the new version of the global_mean bias so I would not worry toooo much

src/aqua/reader/fixer.py
oloapinivad: yes, the problem si that so far whatever data model you specify in the AQUA code will not have effect in any manner. Actually the entire flag must be chgend in the fixer so that data model is true or false, or better pointing to `aqua`

src/aqua/data_model/coordtranformer.py
oloapinivad: yes, there is not test case. @jhardenberg mentioned that he wanted to add it

oloapinivad:
Replied to most of the comments, for this can be merged but please let me know if there anything else you would like to discuss

mnurisso:
Merging!

PR #1869, opened 2025-04-06
Reintroduce check for lat flip

### LINKED ISSUE #1868: Regridding ERA5 to grids with decreasing lats produces flipped image
### machine

Lumi

### catalog

obs

### version

main

### What happened?

If we regrid ERA5 to the F128 grid (a grid with decreasing latitudes) we get a flipped image.
This does not happen if we regrid data from another source like CERES.

This is new and must be due to recent changes, it used to work with a previous version of AQUA (I cannot be specific but about 1 month ago). It is as if we forgot to flip the data as well somewhere. 

```python
from aqua import Reader
readerobs = Reader(catalog='obs', model="ERA5", exp="era5", source="monthly", regrid="F128", rebuild='true')
obs = readerobs.retrieve()
era5=obs['2t']
readerobs.regrid(era5.isel(time=0)).plot()
```

### Error messages

_No response_

None

jhardenberg:
## PR description:

This reintroduces a call to the `flip_lat_dir()` utility before regridding to make sure that in cases where the data model has flipped meridional  coordinates these are reversed again before regridding, to avoid flipped images. 
The call had been removed by mistake in the PR to remove generator support.

Close #1868

----
 - [x] Changelog is updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 Unreleased in the current development version (target v0.14.1):
 
 AQUA core complete list:
+- Fix flipped latitudes when regridding data flipped by datamodel (#1869)
 - Refactor of `plot_single_map()` and `plot_single_map_diff()` functions with projection support (#1854)
 - Refactor time handling: replacement of `datetime` objects and of `pd.Timestamp` lists (#1828)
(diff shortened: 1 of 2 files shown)


jhardenberg:
As discussed https://github.com/DestinE-Climate-DT/AQUA/issues/1868 this will be fixed differently soon, so thi PR should not be merged

jhardenberg:
ok, closing then

PR #1873, opened 2025-04-08
Hovmoller diagnostics refractoring [Ocean3d]

ghosh97:
As per our last discussion on ocean diagnostics refractoring, in this PR I'm going to add only the hovmoller function in a modular way. Now, the goal is to make it up to the netcdf saving.


mnurisso:
Thanks @ghosh97 let me know when a review is needed and also if a proper call should be set to homogeneize the code with the other diagnostics written so far and make your task as smooth as possible.

src/aqua_diagnostics/ocean3d/ocean_drift/hovmoller.py
```diff
@@ -0,0 +1,82 @@
...
+        from ocean3d.ocean_drifts.tools import data_process_by_type
+        from itertools import product
+        import numpy as np
+        
+        if dim_mean is not None:
+            self.data = self.data.mean(dim=dim_mean)
+        data_list = []
+
+        for anomaly, standardise, anomaly_ref in product([False, True], [False, True], ["t0", "tmean"]):
+            data_proc, type, cmap = data_process_by_type(
```
mnurisso, line 64: This data_process_by_type should become part of this class of an utility in this folder, please move it here in the new PR and we will check it as well

mnurisso (commented):
I left a bit of comment, please take a look and let me know if anything requires some clarifications

src/aqua_diagnostics/ocean3d/ocean_drift/hovmoller.py
```diff
@@ -0,0 +1,82 @@
...
+    
+    def run(self):
+        self.retrieve()
+        self.area_select()
+        self.process_data(dim_mean = ["lat","lon"])
+        self.save_netcdf(diagnostic = "Hovmoller", diagnostic_product = "Hovmoller")
+        self.logger.info("Hovmoller diagram saved to netcdf file")
+    
+    def retrieve(self):
+        self.data, self.reader, self.catalog = super()._retrieve(catalog=self.catalog, model=self.model,
```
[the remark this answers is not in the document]
ghosh97: Could you please explain a bit how to use it? Is it in same Diagnostics class?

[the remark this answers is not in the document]
mnurisso: I can show it to you today, please remind me

mnurisso:
Hi @ghosh97 no I'd go step by step, if this is done I'll review it, merge it and we start immediately the new pr, shall I review it now?

mnurisso:
> Yes, please review it. It may lack some parts in stacked data (the data for the Hovmöller plot). I can improve it once I have the Plotting fucntion.

Thanks, I'll do it tomorrow morning. It would be ideal, for the testing and for what is it the final outcome expected from the diagnostic, if a notebook on what is available in this new diagnostic so far is put in `notebooks/diagnostics/ocean3d`.
This can be done eventually in the follow up PR

config/diagnostics/ocean3d/regions.yaml
mnurisso, line 1: We would like to homogeneize these regions file. Tagging again @oloapinivad for a feedback. In timeseries and seaice we do:

```yaml
regions:
  nh:
    longname: North Midlatitudes
    lon_limits: [-180, 180]
    lat_limits: [30, 90]
```

This is very similar to what you're doing but I'd go for a single way to describe regions. A bonus of this implementation will be to have a longname descriptive for the user but a shortcut for the code and configuration files.

src/aqua_diagnostics/ocean3d/ocean_drift/hovmoller.py
```diff
@@ -0,0 +1,131 @@
...
+        logger (Logger): Logger instance for the class.
+        outputdir (str): Directory to save the output files.
+        region (str): Region for area selection.
+        var (list): List of variables to process.
+        stacked_data (xarray.Dataset): Processed data for Hovmoller diagrams.
+    """
+
+    def __init__(
+        self,
+        catalog=None,
```
mnurisso, line 28: Can you specify the type of input? Something like:

```
catalog: str = None, model: str = None, exp: str = None, source: str = None,
regrid: str = None, startdate: str = None, enddate: str = None,
```

src/aqua_diagnostics/ocean3d/ocean_drift/hovmoller.py
```diff
@@ -0,0 +1,131 @@
...
+            exp=exp,
+            source=source,
+            regrid=regrid,
+            startdate=startdate,
+            enddate=enddate,
+            loglevel=loglevel,
+        )
+        self.logger = log_configure(log_name="Hovmoller", log_level=loglevel)
+
+    def run(self, outputdir=None, region=None, var=["thetao", "so"]):
```
mnurisso, line 62: Same as above, this is nothing mandatory but I'm trying to suggest as much as possible longlasting solutions for the maintainance of the code

src/aqua_diagnostics/ocean3d/ocean_drift/hovmoller.py
```diff
@@ -0,0 +1,131 @@
...
+        super().retrieve(var = var)
+        self.logger.info("Data retrieved successfully")
+        self.area_select()
+        self.stacked_data = _data_process_for_drift(
+            data=self.data, dim_mean=["lat", "lon"]
+        )
+        self.save_netcdf(diagnostic="Hovmoller", diagnostic_product="Hovmoller")
+        self.logger.info("Hovmoller diagram saved to netCDF file")
+
+    def retrieve(self):
```
mnurisso, line 84: If you above use the `super().retrieve()` method (which is correct) then you don't need to define another retrieve method here.

src/aqua_diagnostics/ocean3d/ocean_drift/hovmoller.py
```diff
@@ -0,0 +1,131 @@
...
+        # )
+
+    def area_select(self):
+        """
+        Applies area selection to the retrieved data.
+
+        If a region is specified, the data is filtered based on the
+        predefined region's latitude and longitude bounds.
+        """
+        if self.region is not None:
```
mnurisso, line 110: For my understanding, what happens if region is None? You process global data? If so, I'd comment in the run method docstrings that if None you process global data, which can be a useful possibility

src/aqua_diagnostics/ocean3d/ocean_drift/hovmoller.py
```diff
@@ -0,0 +1,131 @@
...
+            lat_s, lat_n, lon_w, lon_e = predefined_regions(self.region)
+            self.data = area_selection(
+                data=self.data, lat=[lat_n, lat_s], lon=[lon_w, lon_e], drop=True
+            )
+
+    def save_netcdf(self, diagnostic, diagnostic_product, rebuild=True):
```
mnurisso, line 116: I like the idea of diagnostic and diagnostic_product not hard-coded (which is something we may want to improve in other diagnostics as well) but could you provide a default value for them?

src/aqua_diagnostics/ocean3d/ocean_drift/util.py
```diff
@@ -0,0 +1,182 @@
+import xarray as xr
+import numpy as np
+from aqua.logger import log_configure
+from itertools import product
+from aqua.util import find_vert_coord, load_yaml, add_pdf_metadata
+import os
+
+
+def process_region_config(regions_yaml):
```
mnurisso, line 9: If we're going to use all the same region file style this kind of function could be standardize between diagnostics

src/aqua_diagnostics/ocean3d/ocean_drift/util.py
```diff
@@ -0,0 +1,182 @@
...
+    Raises:
+        ValueError: If the specified region is not found in the configuration file.
+
+    Notes:
+        - The configuration file is expected to be located at 
+          `../../../../config/diagnostics/ocean3d/regions.yaml` relative to this script.
+        - The configuration file should contain a dictionary with a "regions" key, where 
+          each region is defined with its boundaries (LatN, LatS, LonE, LonW).
+    """
+    current_dir = os.path.dirname(os.path.abspath(__file__))
```
mnurisso, line 51: All this block should use the ConfigPath class from aqua.util module.
Basically you need to:

1. insert the region file in the `src/aqua/cli/diagnostics.py` so that is copied among the others.
2. This will allow aqua install to correctly copy that file in your .aqua folder
3. At this point you can get it from the ConfigPath tool without having to hardcode any path

I can give guidance on this, let me know

src/aqua_diagnostics/ocean3d/ocean_drift/util.py
```diff
@@ -0,0 +1,182 @@
...
+        lon_w = region_boundary.get("LonW")
+    except KeyError:
+        raise ValueError(
+            f"Invalid region name: {region}. Check the region name in config file or update it: {regions_yaml}"
+        )
+
+    return lat_s, lat_n, lon_w, lon_e
+
+
+def _data_process_by_type(**kwargs):
```
mnurisso, line 70: I would be for a refactor of this function. This at the moment hides the Hovmoller class from all the actual computation. I'd insert this as method in the class, called evaluate_something or something similar, to make clear that is the evaluation method.

Can you try to do this?

src/aqua_diagnostics/ocean3d/ocean_drift/util.py
```diff
@@ -0,0 +1,182 @@
...
+    anomaly_ref = kwargs["anomaly_ref"]
+    standardise = kwargs["standardise"]
+
+    process_data = xr.Dataset()
+
+    if anomaly:
+        anomaly_ref = anomaly_ref.lower().replace(" ", "").replace("_", "")
+        if not standardise:
+            if anomaly_ref in ["tmean", "meantime", "timemean"]:
+                cmap = "coolwarm"  # "PuOr"
```
mnurisso, line 98: Also while refactoring this please do not have plot info here since it's still the netcdf creation.

What I suggest is to add an AQUA_attribute_name to the xarray if you then need to identify the different data array produced.

src/aqua_diagnostics/ocean3d/ocean_drift/util.py
```diff
@@ -0,0 +1,182 @@
...
+        cmap = "jet"
+        logger.debug("Data processed for Full values as anomaly = False")
+        type = "Full values"
+
+        process_data = data
+    # logger.debug(f"Data processed for {type}")
+    return process_data, type, cmap
+
+
+def _data_process_for_drift(data, dim_mean: None, loglevel="WARNING"):
```
mnurisso, line 145: As above, this, which wraps the method above, should become a method in the Hovmoller.

mnurisso (commented):
I @ghosh97 I think we're starting to converge to a more robust structure.
I left a bit of comments. The main one is about the functions in the util.py.

What I suggest is to try to refactor them as class methods, maybe simplifying them and removing the plot info which will be added by the second class.
Let me know if you need help for this step.

I think we're coming to a good structure, let me know if I create any doubt.

mnurisso:
Hi @ghosh97 let me know when a further check should be done on my side or if some explanation is necessary

src/aqua_diagnostics/ocean3d/ocean_drift/util.py
ghosh97: what do you mean? I didn't get it.

src/aqua_diagnostics/ocean3d/ocean_drift/util.py
ghosh97: @mnurisso I've updated this in `src/aqua/cli/diagnostics.py`. I think I also have to modify the util function to read the `region.yaml` from the `.aqua` dir.

src/aqua/cli/diagnostic_config.py
```diff
@@ -62,9 +62,9 @@
     ],
     'ocean3d': [
         {
-            'config_file': 'regions.yaml',
-            'source_path': 'diagnostics/ocean3d/config',
-            'target_path': 'diagnostics/ocean3d/config'
+            'region_file': 'regions.yaml',
```
mnurisso, line 65: leave config_file as name here, we will make better than this implementation, but so far config_file is needed as name

src/aqua_diagnostics/ocean3d/ocean_drift/util.py
mnurisso: you can use ConfigPath to get the path and then load_yaml to load the dict, let me know if assistance is needed

src/aqua_diagnostics/ocean3d/ocean_drift/hovmoller.py
```diff
@@ -0,0 +1,180 @@
...
+            anomaly (bool, optional): Specifies whether to compute anomalies. Defaults to False.
+            standardise (bool, optional): Specifies whether to standardize the data. Defaults to False.
+            anomaly_ref (str, optional): Reference for the anomaly computation. Valid options: "t0", "tmean". Defaults to None.
+
+        Returns:
+            process_data (Dataset): Processed data based on the selected preprocessing approach.
+            type (str): Type of preprocessing approach applied
+            cmap (str): Colormap to be used for the plot.
+        """
+        data = kwargs["data"]
```
mnurisso, line 115: Why this use of kwargs? This is quite prone to error if not used as you think. Since they are real argument also mentioned in the docstring I'd simply leave them as real arguments

src/aqua_diagnostics/ocean3d/ocean_drift/hovmoller.py
```diff
@@ -0,0 +1,180 @@
...
+        anomaly = kwargs["anomaly"]
+        anomaly_ref = kwargs["anomaly_ref"]
+        standardise = kwargs["standardise"]
+
+        data = _get_std_anomaly(data, anomaly_ref, standardise, dim = "time")
+
+        return data
+
+
+    def _data_process_for_drift(self, dim_mean: None, loglevel="WARNING"):
```
mnurisso, line 125: If I understood this is the final compute method is this one. If this is the case I'd keep the standard name we're thinking for these methods compute_<something> and keep it not a hidden method

mnurisso (commented):
I left a bit of comments to progress, let me know if everything is clear

ghosh97:
@mnurisso I think this is a bit ready; feel free to do some review. And do you want the tests in this PR?

silviacaprioli (approved):

PR #1877, opened 2025-04-10
Data model into AQUA config

oloapinivad:
## PR description:

This moves the config of the basic AQUA data_model into the config folder.


oloapinivad:
Ok, I make some cleaning, removed the old data model files, and now there is the possibility of loading a different data model yaml file that is installed in the `config` folder. Ready for review!

PR #1883, opened 2025-04-17
improve logging of data_model

oloapinivad:
## PR description:

Small fix to improve the logging and avoid a warning to arise.


mnurisso (approved):
I see that one coverage failed but the motivation seems to be totally negligible so I'd be for a merge.

I wonder if you can include a change here since it is a logger one. In the ConfigPath we have:

`self.logger.warning('No paths found in the main configuration file %s', self.base_available)`

It should not be a warning since the paths in the aqua-config are an extra feature, not the standard behavior. If it's ok to change loglevel here as well I'd then go for a merge

mnurisso:
Merging

PR #1884, opened 2025-04-24
Teleconnection refactor

mnurisso:
## PR description:

WIP draft for the teleconnections refactor. More details will follow

Close #issue_number

----

 - [x] Changelog is updated.
 - [x] environment.yml and pyproject.toml are updated if needed, together with the lumi installation tool. 


src/aqua_diagnostics/teleconnections/base.py
```diff
@@ -0,0 +1,349 @@
...
+            exp (str): The experiment to be used.
+            source (str): The source to be used.
+            regrid (str): The target grid to be used for regridding. If None, no regridding will be done.
+            startdate (str): The start date of the data to be retrieved.
+                             If None, all available data will be retrieved.
+            enddate (str): The end date of the data to be retrieved.
+                           If None, all available data will be retrieved.
+            configdir (str): The directory where the interface file is located.
+                             If None, the default directory will be used.
+            interface (str): The filename of the interface file.
```
oloapinivad, line 34: why do you need an interface file for teleconnections?

src/aqua_diagnostics/teleconnections/README.md
```diff
@@ -1,18 +1,14 @@
 # Teleconnections diagnostic
```
oloapinivad, line 1: does make sense to have a Readme file? we should probably have simply all this information in the documentation, do you agree?

oloapinivad:
It looks very nice! 
I am unable to provide a detailed review on the next, but I ran the notebook and I can leave some small comments

NAO notebook:
- the notebooks use the kwargs structure for simplicity, it would be nice to have a "clean example" with some evident call to the keyword arguments
- the plot class is initialised with the indices, while the correlation are fed into a method. this is a bit counterintuitive to me...
- the comparison of ncar in the very end, do you have any clue why we have less variability? due to the 3 month average?
- the save_netcdf is not shown in this notebook, I guess is available as for ENSO

ENSO notebook
- I tried to plot without computing the index before, the tool failed with a non very informative error while plotting and not while initiliazing

concordance map
- I suspect the definition of "concordance" is a bit arbitrary. what about "statistical validation" or "bootstrapping"?
- the notebook is still using the old class scheme, correct? Teleconnection, evaluate_index, etc...

MJO notebook
- there is no documentation at all here

src/aqua/util/sci_util.py
```diff
@@ -253,6 +253,11 @@ def select_season(xr_data: xr.DataArray or xr.Dataset, season: str):
...
         selected_months = triplet_months[season]
-        return xr_data.sel(time=(xr_data['time.month'] == selected_months[0]) | (xr_data['time.month'] == selected_months[1]) | (xr_data['time.month'] == selected_months[2]))
+        selected =  xr_data.sel(time=(xr_data['time.month'] == selected_months[0]) | (xr_data['time.month'] == selected_months[1]) | (xr_data['time.month'] == selected_months[2]))
+        # Add AQUA_season attribute
+        selected.attrs['AQUA_season'] = season
+        return selected
+    elif season == 'annual':
+        return xr_data
     else:
         raise ValueError("Invalid season abbreviation. Please use one of the provided abbreviations.")
```
oloapinivad, line 263: provided by whom?

src/aqua_diagnostics/teleconnections/base.py
mnurisso: You have defined the ENSO region, the NAO station points, the variable to use for the index (so that I'm cmor ready) and for MJO I have other options (like for olr if I have to flip the sign, cmor to grib if I recall correctly is different).

src/aqua_diagnostics/teleconnections/README.md
mnurisso: It's a remnant of the old diagnostic structure, I'll remove it

src/aqua_diagnostics/teleconnections/base.py
```diff
@@ -0,0 +1,349 @@
...
+
+    Args:
+        maps (list): The list of maps to be homogenized.
+        ref_maps (list): The list of reference maps to be homogenized.
+        var (str): The variable to be used. Default is None.
+
+    Returns:
+        tuple: The homogenized maps and reference maps.
+    """
+    if var == 'msl':
```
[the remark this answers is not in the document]
oloapinivad: would make sense to do something "if unit is Pa then convert to hPa"?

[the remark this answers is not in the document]
mnurisso: Sure, I'll update it like this

src/aqua_diagnostics/teleconnections/base.py
mnurisso: sure I'll change it

src/aqua/util/sci_util.py
mnurisso: done

oloapinivad:
I see your point, I think it makes sense to keep it as it is! Thanks for answering to all the other comments!

mnurisso:
> the comparison of ncar in the very end, do you have any clue why we have less variability? due to the 3 month average?

No I'm not sure of the cause and I remember we already discussed this. What I checked now was simply to have the same results as before the PR.

mnurisso:
Everything is addressed.

The boostrap is now aligned to the new classes and work (the bootstrap cli not yet, if needed in the future is a quick task).

With @jhardenberg we're discussing if the file names should be nao.index.... or teleconnections.nao_index... 
I personally don't like to have to build a complex diagnostic_product, what's your opinion on this?
I would be also ok to have the grouping splitting the different teleconnections in the dashboard in the future. It will also make more evident if a new teleconnections will be added (e.g MJO).

oloapinivad (approved):
I did not check anything but 59 files it is enough and I blindly trust you.

PR #1879, opened 2025-04-11
Lra level selection new feature

mnurisso:
## PR description:

As asked in #1878 we introduce the possibility to select a level or list when producing the LRA.

## Issues closed by this pull request:

Close #1878 

----

Please leave the checkboxes that apply to this pull request.
If you find a missing checkbox, please add it to the list.
Make sure to complete the checkboxes before applying the "ready to merge" label.
Please apply the "run tests" label if you want to trigger CI tests.

 - [ ] Tests are included if a new feature is included.
 - [ ] Documentation is included if a new feature is included.
 - [ ] Docstrings are updated if needed.
 - [ ] Changelog is updated.
 - [ ] Notebooks which requires changes are updated. 
 - [ ] environment.yml and pyproject.toml are updated if needed, together with the lumi installation tool. 
 - [ ] Diagnostic config files path are updated in the console if needed

PR #1886, opened 2025-04-27
Expand LRA to compute also other statistics and some other refinements

### LINKED ISSUE #1885: LRA generate also other statistics
### Feature request

It would be useful to extend the LRA to be a general 'data extractor' capable of deriving also other statistics such as daily maxima or monthly standard deviation etc. BAsically the idea would be to allow specifying also other stats for the TimStat method in addition to the mean.

jhardenberg:
## PR description:

This expands LRA capabilities to compute also additional statistics, not only the mean.
It contains also some other minor improvements to LRA functionality, in particular grouping annual files using cdo (which is significantly faster than using xarray). 

Close #1885

----

 - [x] Changelog is updated.

### DIFF
--- a/src/aqua/lra_generator/lra_generator.py
+++ b/src/aqua/lra_generator/lra_generator.py
@@ -47,7 +47,6 @@ class LRAgenerator():
                  exclude_incomplete=False,
                  stat="mean",
                  compact="xarray",
-                 cdo="cdo",
                  cdo_options=["-f", "nc4", "-z", "zip_1"],
                  **kwargs):
         """
@@ -90,7 +89,6 @@ class LRAgenerator():
             stat (string, opt):      Statistic to compute. Can be 'mean', 'std', 'max', 'min'.
             compact (string, opt):   Compact the data into yearly files using xarray or cdo.
                                      If set to None, no compacting is performed. Default is "xarray"
-            cdo (string, opt):       Path to the cdo executable used for compacting, default is "cdo"
             cdo_options (list, opt): List of options to be passed to cdo, default is ["-f", "nc4", "-z", "zip_1"]
             **kwargs:                kwargs to be sent to the Reader, as 'zoom' or 'realization'
                                      please notice that realization will change the file name 
@@ -180,7 +178,6 @@ class LRAgenerator():
         if self.compact not in ['xarray', 'cdo', None]:
             raise KeyError('Please specify a valid compact method: xarray, cdo or None.')
         
-        self.cdo = cdo
         self.cdo_options = cdo_options
         if not isinstance(self.cdo_options, list):
             raise TypeError('cdo_options must be a list.')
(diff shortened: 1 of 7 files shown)


jhardenberg:
The first addition is a new `compact` keyword to `LRAgenerator` which allows specifying `cdo` as an alternative method to create the yearly files. In my tests trying to extract data for a limited area at 0.05 degrees, it was taking 2 minutes to generate each month and the above 50min to compact into a yearly file! Now with cdo this takes only one and half minutes. The options for cdo are by default to create a nc4 zipped level 1 file, but they are exposed with another keyword. To be decided if not to make 'cdo' the default option. Currently it is possible to choose 'cdo', 'xarray' or None (which would avoid running the compacting of the monthly in yearly files.

jhardenberg:
next change is the 'stat' keyword  (default is 'mean') . The output filenames are like "2t_ssp370_r005_daily_min_myregion_202001.nc"

jhardenberg:
One important point: since this chanegs the output naming, if this PR  is merged it will trigger a complete rebuild of all LRAs we have ...maybe we should omit 'mean' in the output filenames if the statistic is 'mean' ?  Or maybe we should rename by hand all output files before the next time LRA is run?

jhardenberg:
The new option to drop empty gridpoints for a regional selection is called 'drop' and the one to decide if and how to do the yearly compacting of files is called 'compact' .... suggestions welcome if you have better names

mnurisso:
Hi, I think we need to discuss this in the next in person meeting in order to be sure we do not only patch the LRA but we have a more organic update of the structure accomodating all the features. I'd like to have also the level selection from #1879 in this list of new features.

I'd also like to have both the possibility of having the kwargs as cli args but also in the yaml file. Maybe we need to think which structure this file should have to accommodate the features.

mnurisso (approved):

jhardenberg:
Ok, last changes which we agreed: 
* I implemented the new way to do 'drop' in the regional area selection. Tested in a workflow and works like a charm, producing nice and smaller files only for the region of interest. Possibly because it takes less to write the also now it takes just 1 minute to process one month of climatedt1 ICON ssp370 data regridded at 0.05 degrees
* removed the 'cdo' option but left 'cdo_options' (we will think again about this when we decide how to further change LRA)

src/aqua/lra_generator/lra_generator.py
```diff
@@ -39,11 +40,15 @@ def __init__(self,
...
+                 region=None, drop=False,
                  overwrite=False, definitive=False,
                  performance_reporting=False,
                  rebuild=False,
-                 exclude_incomplete=False, **kwargs):
+                 exclude_incomplete=False,
+                 stat="mean",
+                 compact="xarray",
+                 cdo_options=["-f", "nc4", "-z", "zip_1"],
+                 **kwargs):
```
oloapinivad, line 51: what is `zip_1`? why not simply a`zip`?

jhardenberg: zip_1 is zip compression level 1
We use the same option when writing with xarray

oloapinivad: I suspect it is the same of using `zip`! but of course no hassle, go on and merge for me

jhardenberg:
All works, tests passing, merging

PR #1893, opened 2025-05-05
Polytope support

### LINKED ISSUE #1534: Polytope support
### Feature request

We want to explore if the intake-gsv module can open data also with the GSVRetrieve making use of the `engine='polytope'`. For this reason @ghosh97 will investigate what @igonzal-bsc already tested with the GSV, with the autentication and if AQUA breaks somehow implementing this.

jhardenberg:
## PR description:

This adds polytope support with the new `engine` keyword (can be 'fdb' (default) or 'polytope'). The keyword can either be added to a catalog entry or be specified directly in the reader (it exploits the fact that the kwargs are passed directly to intake).

To be tested for speed and if remote functionality really works well. Please notice that the current implementation in the GSVRetriever uses a temporary grib file on disk for each message, so it has to be seen if this is ok in terms of speed.

The user will need credentials to access polytope. To this end get polytope credentials and create a `~/.polytopeapirc` file in your home with this content:

```
{
    "user_email" : "<your.mail>",
    "user_key" : "<your.key>"
}
```

I will add also some info on this in the docs.

Close #1534 

----

 - [x] Changelog is updated.

### DIFF
--- a/.github/workflows/aqua.yml
+++ b/.github/workflows/aqua.yml
@@ -145,9 +145,10 @@ jobs:
         run: |
           # Initialize the AQUA catalog in the $HOME/.aqua folder
           aqua -vv install github
+          # Add the climatedt-phase1 catalog (for polytope testing)
+          aqua -vv add climatedt-phase1
           # Add the ci/cd catalog
           aqua -vv add ci
-          aqua -vv add climatedt-phase1
           # Double check the catalogs
           aqua -vv list
           # add default paths to the config
--- a/.github/workflows/aqua.yml
+++ b/.github/workflows/aqua.yml
@@ -151,7 +151,7 @@ jobs:
           # Double check the catalogs
           aqua -vv list
           # add default paths to the config
-          cat >> $HOME/.aqua/config.yaml <<EOF
+          cat >> $HOME/.aqua/config-aqua.yaml <<EOF
           paths:
               grids: ./AQUA_tests/grids
               weights: ./AQUA_tests/weights
--- a/.github/workflows/aqua.yml
+++ b/.github/workflows/aqua.yml
@@ -157,6 +157,13 @@ jobs:
               weights: ./AQUA_tests/weights
               areas: ./AQUA_tests/weights
           EOF
+          # Set up polytope secrets
+          cat >$HOME/.polytopeapirc << EOF
+          {
+            "user_email": "${{ secrets.POLYTOPE_MAIL }}",
+            "user_key": "${{ secrets.POLYTOPE_KEY }}"
+          }
+          EOF
       - name: Run tests
         run: |
           # Produce the XML report
--- a/.github/workflows/aqua.yml
+++ b/.github/workflows/aqua.yml
@@ -147,8 +147,16 @@ jobs:
           aqua -vv install github
           # Add the ci/cd catalog
           aqua -vv add ci
+          aqua -vv add climatedt-phase1
           # Double check the catalogs
           aqua -vv list
+          # add default paths to the config
+          cat >> $HOME/.aqua/config.yaml <<EOF
+          paths:
+              grids: ./AQUA_tests/grids
+              weights: ./AQUA_tests/weights
+              areas: ./AQUA_tests/weights
+          EOF
       - name: Run tests
         run: |
           # Produce the XML report
--- a/docs/sphinx/source/advanced_topics.rst
+++ b/docs/sphinx/source/advanced_topics.rst
@@ -198,6 +198,26 @@ levels to use for each chunk.
 
 .. _lev-selection-regrid:
 
+Polytope access to Destination Earth data
+-----------------------------------------
+
+It is poossible to access ClimateDT data availble on the 'Databridge' for the DestinE ClimateDT also remotely, from other machines,
+using the 'polytope' access. to this end you will need to specify ``engine="polytope"`` when instantiating the `Reader` or adding
+the argument ``engine="polytope"`` as an additional argument in the intake data source.
+
+.. code-block:: python
+
+    reader = Reader(model="IFS-NEMO", exp="ssp370", source="hourly-hpz7-atm2d", engine="polytope")
+    data = reader.retrieve(var='2t')
+
+This allows accessing ClimateDT data on the Databridge also remotely from other machines.
+
+In order for this to work you will need to store an access token in the file ``~/.polytopeapirc`` in your home directory.
+Please follow the instructions in the `Polytope documentation <https://github.com/destination-earth-digital-twins/polytope-examples>`_ to 
+set up appropriate credential and create this file. 
+As a preliminary step, you will need to be registered on the `Destine Service Platform  <https://auth.destine.eu/>`_ 
+and have requested "upgraded access" to the data.
+
 Level selection and regridding
 ------------------------------
 
--- a/docs/sphinx/source/advanced_topics.rst
+++ b/docs/sphinx/source/advanced_topics.rst
@@ -212,21 +212,21 @@ the argument ``engine: polytope`` as an additional argument in the intake catalo
 
 This allows accessing ClimateDT data on the Databridge also remotely from other machines.
 
-To access Destination Earth ClimateDT data you will need to be registered on the `Destine Service Platform  <https://platform.destine.eu/>`_ 
+To access Destination Earth ClimateDT data you will need to be registered on the `Destine Service Platform  <https://platform.destine.eu/>`_
 and have requested "upgraded access" to the data (follow the link "Access policy upgrade" under your username at the top left corner of the page).
 
-In order for this to work you will need to store an access token in the file ``~/.polytopeapirc`` in your home directory.
+In order for his to work you will need to store an access token in the file ``~/.polytopeapirc`` in your home directory.
 You can create this file following two proceures:
 
-1. ** Using DestinE Service Platform credentials **: 
+1. **Using DestinE Service Platform credentials**: 
 
-Follow the instructions in the `Polytope documentation <https://github.com/destination-earth-digital-twins/polytope-examples>`_ 
+Follow the instructions in the `Polytope documentation <https://github.com/destination-earth-digital-twins/polytope-examples>`_
 and the username and password which you defined for the Destine Service Platform to download the credentials into this file. 
 
-2. ** Using ECMWF credentials **:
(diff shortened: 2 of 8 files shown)


jhardenberg:
Example on our remote machine Mafalda, accessing climatedt-phase1 data through polytope:

image
image

mnurisso:
Hi, I tried myself and I succeeded with a couple of differences I'd like to discuss:

First, I had to comment in the intake_gsv.py the lines which check if FDB_HOME exists as a path, this should not be done in the engine='polytope' case.

Then my ~/.polytopeapirc file contains only the key, if I add manually the email it does not work.

Third I'd find a way to well document the key generation, which is unfortunately not trivial.

jhardenberg:
Yes, I confirm the issue with checking FDB_HOME paths ..... 
I tried to use a simple 'do it only if engine=='fdb' but unfortunately for some weird reason the entire class is initialized TWICE by intake, once with and once without kwargs (so the first time it runs engine=='fdb' (the default) .... 
I will try and find a solution

jhardenberg:
The fact that you cannot put your email in teh credentials is weird, it is my understanding that the email is basically the 'login' and part of the credentials

mnurisso:
> The fact that you cannot put your email in teh credentials is weird, it is my understanding that the email is basically the 'login' and part of the credentials

What I did was to use some other script to generate the key and then run aqua in another notebook. It's for sure something to explore more, maybe the email was cached somehow (it was done in the same vscode instance)

jhardenberg:
Ok, I really studied the issue: intake v1 really insists on initializing the driver clas twice: onc ebecause we access the source (without any arguments or kwargs) and then again becuase we specify kwargs. I did not find a way to avoid it, but  the current solution prevents it from checking fdb paths the first time and it will do so the second time only if not engine=='polytope'. This way now it all works well.

jhardenberg:
I added a tentative description in the docs on how to obtain the polytope credentials. Please be free to change/integrate

jhardenberg:
@oloapinivad @mnurisso maybe we could add a test which uses polytope to download a single frame from the ClimateDT. 
In that case should we install also the climatedt-phase1 catalogue when doing the tests or rather add a specialized entry to the CI catalogue?

mnurisso:
> @oloapinivad @mnurisso maybe we could add a test which uses polytope to download a single frame from the ClimateDT. In that case should we install also the climatedt-phase1 catalogue when doing the tests or rather add a specialized entry to the CI catalogue?

Sure that's the plan in my opinion. To have a secret with a polytope key and test it

jhardenberg:
what about catalog to use? Just one entry copied from the climatedt catalogue added to the CI catalogue or do we install the full climatedt-phase1 catalogue also for the tests?

mnurisso:
> what about catalog to use? Just one entry copied from the climatedt catalogue added to the CI catalogue or do we install the full climatedt-phase1 catalogue also for the tests?

I'd go for the climatedt-phase1 (with the idea of updating the catalog to the latest O cycle)

oloapinivad:
I agree, the full ClimateDT-phase1 requires less maintenance and in principle we could also test STAC access on it without doing anything specific.

jhardenberg:
Ok, so what we would need to do , so that it can find the grids when it is using the climatedt-phase1 catalog is to add this in the `climatedt-phase1/machine.yaml` file, correct?
```
ci:
  paths:
    grids: ./AQUA_tests/grids
    weights: ./AQUA_tests/weights
    areas: ./AQUA_tests/weights
  intake:
    TEST_PATH: ./AQUA_tests/models
    FDB_HOME: /app
```

jhardenberg:
I am using a possibly better solution: I add those paths directly as a default  to `config-aqua.yaml` in the test workflow. This way we are independent from keeping a 'ci' machine in all catalogues.

jhardenberg:
tests added and working!

jhardenberg:
I had to change again how I deal with the 'checking fdb paths issue'. This version works in all cases but it feels like a hack.

jhardenberg:
On the solution which I implemented to avoid the 'double initialization' issue: there would be different solutios (like moving the construction of the time axis and checking the  paths to the to_dask method) but actually maybe this is not worth it: we would like anyway to rewrite everything to adapt it to intake 2.0 .... so it is useless to put too much effort in refactoring a intake v1 version

.github/workflows/aqua.yml
```diff
@@ -145,10 +145,26 @@ jobs:
...
           # Initialize the AQUA catalog in the $HOME/.aqua folder
           aqua -vv install github
+          # Add the climatedt-phase1 catalog (for polytope testing)
+          aqua -vv add climatedt-phase1
           # Add the ci/cd catalog
           aqua -vv add ci
           # Double check the catalogs
           aqua -vv list
+          # add default paths to the config
+          cat >> $HOME/.aqua/config-aqua.yaml <<EOF
```
mnurisso, line 155: Maybe it is now fine to have the paths added like this, but a proper aqua console command should be added. This way the users on a new machine will have a clear command to specify where to store grids and other auxiliary files.

I'd leave it as it is here and open an issue on this topic only.

docs/sphinx/source/advanced_topics.rst
```diff
@@ -198,6 +198,26 @@ levels to use for each chunk.
 
 .. _lev-selection-regrid:
 
+Polytope access to Destination Earth data
+-----------------------------------------
+
+It is poossible to access ClimateDT data availble on the 'Databridge' for the DestinE ClimateDT also remotely, from other machines,
```
mnurisso, line 204: typo in possible

docs/sphinx/source/advanced_topics.rst
```diff
@@ -198,6 +198,26 @@ levels to use for each chunk.
...
+using the 'polytope' access. to this end you will need to specify ``engine="polytope"`` when instantiating the `Reader` or adding
+the argument ``engine="polytope"`` as an additional argument in the intake data source.
```
mnurisso, line 206: What do you mean? In the catalog entry yaml file? This is cool if that's the case, nice!

docs/sphinx/source/advanced_topics.rst
```diff
@@ -198,6 +198,26 @@ levels to use for each chunk.
...
+    reader = Reader(model="IFS-NEMO", exp="ssp370", source="hourly-hpz7-atm2d", engine="polytope")
+    data = reader.retrieve(var='2t')
+
+This allows accessing ClimateDT data on the Databridge also remotely from other machines.
+
+In order for this to work you will need to store an access token in the file ``~/.polytopeapirc`` in your home directory.
+Please follow the instructions in the `Polytope documentation <https://github.com/destination-earth-digital-twins/polytope-examples>`_ to 
+set up appropriate credential and create this file. 
+As a preliminary step, you will need to be registered on the `Destine Service Platform  <https://auth.destine.eu/>`_ 
+and have requested "upgraded access" to the data.
```
mnurisso, line 219: Is there a link to explain how to ask upgraded access?

src/aqua/reader/reader.py
```diff
@@ -75,7 +75,7 @@ def __init__(self, model=None, exp=None, source=None, catalog=None,
             preproc (function, optional): a function to be applied to the dataset when retrieved. Defaults to None.
             convention (str, optional): convention to be used for reading data. Defaults to 'eccodes'.
                                         (Only one supported so far)
-            **kwargs: Arbitrary keyword arguments to be passed as parameters to the catalog entry.
+            **kwargs: Arbitrary keyword arguments to be passed as additional parameters to the catalog entry.
```
mnurisso, line 78: I'm thinking how to make this Docstring clear.
I know you can create not only Args and Returns blocks, but also one for the Kwargs. This would be ideal, because the info on zoom and engine can be better explained there, I think the Reader docstring should be very clear since these are the ones that any user will check when in doubt.

mnurisso (approved):
I left a bit of comments but I'm ok with the PR.
About the hack and the consideration on intake 2 I totally agree. When updating a refactor will be needed, no need to become crazy now with something that will be changed later in the future

docs/sphinx/source/advanced_topics.rst
jhardenberg: Yes, I rephrased a bit to make it clearer

docs/sphinx/source/advanced_topics.rst
jhardenberg: I added some more explanation on how to do that

src/aqua/reader/reader.py
jhardenberg: done. @mnurisso please counter-check if you like the changes and then we can proceed with merging!

docs/sphinx/source/advanced_topics.rst
```diff
@@ -198,6 +198,35 @@ levels to use for each chunk.
...
+
+This allows accessing ClimateDT data on the Databridge also remotely from other machines.
+
+In order for this to work you will need to store an access token in the file ``~/.polytopeapirc`` in your home directory.
+Please follow the instructions in the `Polytope documentation <https://github.com/destination-earth-digital-twins/polytope-examples>`_ to 
+set up appropriate credential and create this file. 
+A sample file will look like this:
+
+.. code-block:: text
+
```
oloapinivad, line 221: just tested this, and it seem that adding the email breaks the login. For me it worked only with the user key

jhardenberg: I tried: if I remove the email from`.polytopeapirc` the this breaks login for me.
My guess is that maybe you are not providing the right email for the login?
Possibly without email specified it guesses it somehow and maybe in your case it gets it right for some reason ?
Or gets it from some environment variable?
I will PM you sending my own credentials, so you can try if using mine specifying email and token works

jhardenberg: It turns out that the email and token which you need to use are the same which can be found also in your ECMWF account pages. Login to www.ecmwf.int and then go to [https://api.ecmwf.int/v1/key/](https://api.ecmwf.int/v1/key/)

Are these email and token which you were using?

jhardenberg:
I further improved the documentation to better explain where to find the contents for `.polytopeapirc`

docs/sphinx/source/advanced_topics.rst
jhardenberg: I added the ultimate description in the docs for this now

PR #1895, opened 2025-05-06
Bridge STAC API implementation to guess Bridge availability

### LINKED ISSUE #1894: STAC API to get info on what is on the bridge
### Feature request

Talking with Sebastian Milinksi we found out there is a STAC API https://climate-catalogue.lumi.apps.dte.destination-earth.eu/ that can be used to interrogate the Data Bridge and see what is currently available. 

I did a test today and it is actually possible to get the start and end dates of the data available from the bridge in a pretty straightforward way. Here below you can see a simple function which implements this. 

I think this can be put inside our intakeGSV driver, and getting the request parameter from our ClimateDT catalog we can develop a new way to define where the data is. A bit of caution must be only done considering the variables.

Please notice that for now the API cover only the first five years. And - following Sebastian info - updates are triggered manually: this should happen automatically on a schedule or triggered by the workflow in the medium term. 

```python
import requests

# Define your parameters
params = {
    "activity": "scenariomip",
    "class": "d1",
    "dataset": "climate-dt",
    "experiment": "ssp3-7.0",
    "expver": "0001",
    "generation": "1",
    "model": "icon",
    "realization": "1",
    "stream": "clte",
    "levtype": "sfc",
    "resolution": "standard",
    "type": "fc",
    "param": "260048",
}

# (Replace this with the actual API endpoint if available)
BRIDGE_API_URL = "https://climate-catalogue.lumi.apps.dte.destination-earth.eu/api/stac"
def get_stac_bridge_dates(params, base_url=BRIDGE_API_URL):
    """
    Function to get from the STAC data bridge the available
    dates of a dataset on the bridge
    
    Args:
        params (dict): Dictionary of parameters to interrogate the STAC API.
                       In principle, the same as the usual FDB request
        base_url (str): URL for the STAC API

    Returns:
        tuple: A tuple containing the start and end dates of the dataset
    """

    # Define the base URL for the STAC API
    params['root'] = 'root'
    response = requests.get(base_url, params=params)
    stac_json = response.json()
    check = stac_json['links'][0]['title']
    if check != 'date':
        raise ValueError(f"The first link in the response is not a date link, but {check}")

    # specific extraction of the dates
    dates = next(link['generalized_datacube:dimension']['values']
                for link in stac_json['links']
                if 'generalized_datacube:dimension' in link)
    sorted_dates = sorted(dates)
    
    return sorted_dates[0], sorted_dates[-1]

get_stac_bridge_dates(params)
```

Combined with polytope access in #1893 in principle this can allow anyone fron anywhere to access ClimateDT data.
Potentially, we can write github tests which access data from the bridge seamlessly

oloapinivad:
## PR description:

This follows what discussed in #1894, showing an initial implementation that can possibly cleaned up and refined. 
Currently tested with storylines with the following code. You have to set in your catalog `bridge_end_date: stac`.

```python
from aqua import Reader
reader = Reader(model="IFS-FESOM", exp="story-2017-control", source="daily-hpz7-oce3d", loglevel='debug')
data = reader.retrieve()
data
```

If we like this implementation, I could modify the catalog for storylines at least and push ECMWF to have this stac updated. 

Close #1894 

----

 - [ ] Changelog is updated.


mnurisso:
> If we like this implementation, I could modify the catalog for storylines at least and push ECMWF to have this stac updated.

Nice!
Storylines btw have for the T2K the fixed data on the HPC, so at least that source I'd keep it as it is

PR #1937, opened 2025-05-16
Hovmoller review

mnurisso:
## PR description:

Since this is the first chunck to be added I spent a bit more time designing how the final look should be.
The PR was already at a good stage but you can see that:

1. always create a notebook to help the reviewer and users
2. docstrings with the same style
3. more info in the comments and data type
4. more linting
5. I also checked and updated some small malfunctioning

In the next steps I'll focus only on the last points, having this as good example of the final look.
Major: the files are now under ocean_drift, since we're going in the direction of splitting the diagnostics to small chuncks to simplify the maintainance

Review of #1873


ghosh97:
Thanks for the PR, I'll keep in mind these parts. 
And I think we can merge this PR, I've tested the notebook also.

PR #1944, opened 2025-05-21
Fixes for ICON `o001` run

oloapinivad:
## PR description:

Following what has been discussed in https://jira.eduuni.fi/browse/CSCDESTINCLIMADT-660 I introduce two small fixes for the ICON esuite run o000. 
This triggers the decumulation for wind stresses and the correct unit for cloud cover.

Of course, this is not enough, since we need to 
- Port this change to the operational branch, and make a new a release
- Update the catalog generator (and therefore the branch) so that the fix is used actively.

### DIFF
--- a/config/fixes/climatedt-phase2.yaml
+++ b/config/fixes/climatedt-phase2.yaml
@@ -21,3 +21,14 @@ fixer_name:
         vars:
             tcc:
                 src_units: frac
+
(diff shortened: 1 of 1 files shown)


mnurisso:
Issue will be solved with the ICON scripts, this is not needed luckily, closing.

PR #1947, opened 2025-05-22
Add regrid option in aqua_analysis

### LINKED ISSUE #1946: Regrid possibility in aqua_analysis
### Feature request

It would be useful to have the possibility in aqua_analysis to regrid input data, since we're now using AQUA on some models for which we regrid data.

silviacaprioli:
## PR description:
This PR adds the possibility in aqua_analysis to regrid input data.

Close #1946

----

 - [ ] Changelog is updated.

### DIFF
--- a/cli/aqua-analysis/aqua-analysis.py
+++ b/cli/aqua-analysis/aqua-analysis.py
@@ -202,6 +202,7 @@ def main():
     model = args.model or config.get('job', {}).get('model')
     exp = args.exp or config.get('job', {}).get('exp')
     source = args.source or config.get('job', {}).get('source', 'lra-r100-monthly')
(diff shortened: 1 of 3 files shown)


mnurisso:
Thanks! I'd discuss this more in details the next monday. I think not all the diagnostics take the regrid argument now, so that we may need to keep this PR open until they're regrid ready, but @jhardenberg knows probably more than me about this.

jhardenberg:
For now this adds the regrid option only to the cli_checker, right? It does not actually pass the regrid option to any diagnotics....

PS: docs will also need to be updated

jhardenberg:
It was already possible to regrid data by passing the regrid option explicitly to every diagnostic in the aqua_analysis confiog file. The idea of this PR is to specify this only once, occerct?

Issue #1938, opened 2025-05-19
[OPERATIONAL] Running thread for tracking AQUA and AQUA-web issues


silviacaprioli:
As reported by @mnurisso  and @oloapinivad  there are few issues concerning automatic figure generation in operational setup:

- [x] **Missing automatic webhook updates for aqua-web operational**
The webhook for triggering updates in aqua-web operational is not functioning automatically. As a result, ICON plots are not being updated regularly. -> Frontend responsability, not an AQUA issue

- [x] **Missing ecmean in ICON figures**
The ecmean field is missing in the ICON-related plots. This might be due to an issue with the area file.

- [x] **IFS-NEMO figures missing from the lumi-o bucket**
The IFS-NEMO figures are not showing up in the lumi-o bucket. They may have been generated correctly on the virtual machine, but something might have gone wrong during the push process. Related issue: https://jira.eduuni.fi/browse/CSCDESTINCLIMADT-645

- [x] **Missing ecmean in IFS-NEMO figures**
The ecmean field is missing in the IFS-NEMO-related plots. This might be due to an issue with the area file.
A discussion is ongoing to find a solution for the file generation. https://jira.eduuni.fi/browse/CSCDESTINCLIMADT-658

- [x] **Fixes problem**
ECmean has problem with wind stresses, and total cloud cover is stored in percentage instead of fraction: 
aqua PR (probably not needed): https://github.com/DestinE-Climate-DT/AQUA/pull/1944 
workflow issue: https://earth.bsc.es/gitlab/digital-twins/de_340-2/workflow/-/issues/1079
jira ticket: https://jira.eduuni.fi/browse/CSCDESTINCLIMADT-660
Addressed on the ICON side in the workflow issue

- [x] **Biases and teleconnections IFS-NEMO**
The diagnostic did fail in horizontal biases due to: `Error processing tprate: <urlopen error [Errno 101] Network is unreachable>`. To be investigated @silviacaprioli. 
Logfile: `/appl/AS/AUTOSUBMIT_DATA/o005/aqua-analysis/climatedt-e25.1/IFS-NEMO/o005` (in the VM)
Jira ticket: https://jira.eduuni.fi/browse/CSCDESTINCLIMADT-661 #1954 

- [x] **experiments.yaml not uploaded**
 This file is explicitly requested by the dashboard (it is not used by aqua-web). It is created and maintained by make_contents.py but then push_analysis.sh does not actually upload it !  Addessed in #1951 

[...]

PR #1950, opened 2025-05-26
Update experiments.yaml file when pushing to dashboard

jhardenberg:
## PR description:

This makes sure that the file `experiments.yaml` which is needed by the dashboard is correctly updated by `push_analysis.sh` and then pushed back to the LUMI-O bucket.
Incidentally this required adding a functionality to `push_s3.py` allowing it also to GET a file besides pushing.

This addresses the last point of https://github.com/DestinE-Climate-DT/AQUA/issues/1938 . We will need to open a 
separate PR for v0.13 for this really to be a solution for operations.

----
 - [x] Changelog is updated.


mnurisso (approved):
Merging

Issue #1972, opened 2025-06-03
Wrong time axis in timeseries in containers


oloapinivad:
### machine

Levante

### catalog

operational

### version

main

### What happened?

We are noticing that plots in the timeseries is wrong by one year, showing  year 1989 instead of year 1990 as in the figure below

![Image](https://github.com/user-attachments/assets/1382bb93-e7e7-4694-8836-ef8ec7ff6929)

Our assumption is that there is a subtle mismatch in time zones which shows up in the container only, which makes xarray/numpy/pandas roll back the time axis by two hours on Lumi and by one hour in MN5. This is a problem because matplotlib by default floor the year, so that even if it is 1989-12-31T22 it reads as 1989. 

So far we did not understand how to fix this by setting up the right binding.

### Code to reproduce the issue

```python
#!/usr/bin/env python3

import xarray as xr
import datetime
import matplotlib.pyplot as plt
import matplotlib.dates as mdates
import pandas as pd

# define paths
folder = '/projappl/project_462000911/mnurisso/AQUA/netcdf'
data_annual = 'timeseries.timeseries.ci.ERA5.era5-hpz3.2t.frequency_annual.nc'
data_monthly = 'timeseries.timeseries.ci.ERA5.era5-hpz3.2t.frequency_monthly.nc'

# Load the data
ds_annual = xr.open_dataset(f'{folder}/{data_annual}')
ds_monthly = xr.open_dataset(f'{folder}/{data_monthly}')

ds_annual = ds_annual['2t']
ds_monthly = ds_monthly['2t']

print(ds_monthly.time)
print(ds_annual.time)

fig, ax = plt.subplots(figsize=(10, 6))

# Plot the timeseries
ds_monthly.plot(ax=ax)
ds_annual.plot(ax=ax)

# Save the plot
plt.savefig('2t_timeseries.png')
```

### Are you interested in making a pull request?

None

oloapinivad:
Just found out that setting 

```python
import matplotlib as mpl
mpl.rcParams['timezone'] = 'EET'
```

solves the problem for the container. So my speculation is now that the internal engine of matplotlib to convert from numbers to dates (e.g. `(mdates.num2date(ax.get_xticks()[0])`) inside the container is not recognizing correctly that we are in EET and assumes we are in UTC. So it convert all the time by two hours unless we specify this

oloapinivad:
[...]

PR #1975, opened 2025-06-04
possible fix for UTC

oloapinivad:
## PR description:

I think I am close to solution of #1972. The container is inheriting the time zone from the machine but believes this is UTC, so that it creates an artificial offset. With this `no-mount` option, which we need to check with HPC admin if is the best way to go, this is blocked and so the container time zone works as expected.

Close #i1972

## People involved:

@mnurisso @TracyMcBean 

----

 - [ ] Changelog is updated.

### DIFF
--- a/cli/aqua-container/load-aqua-container.sh
+++ b/cli/aqua-container/load-aqua-container.sh
@@ -303,7 +303,7 @@ done
 echo "Perfect! Now it's time to ride with AQUA ⛵"
 
 #echo "singularity $cmd --cleanenv $env_args $bind_args $AQUA_container $script"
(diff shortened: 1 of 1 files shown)


TracyMcBean:
Hi @oloapinivad that would be great if adding `no-mount` solves this. This fix looks like it will have to go into the workflow where the aqua container is called or?

oloapinivad:
Exactly, if you want to give it a try this should it is were singularity call the aqua container!

oloapinivad:
Following discussion with @kinow, it seems that this is a Singularity bug. I would merge the current hotfix so this does not show up for now. I would put it only in the main so far, to prevent hiccups in the operational

mnurisso (approved):

PR #1997, opened 2025-06-06
Raise an error if retrieve() in core diagnostic returns an empty dataset

silviacaprioli:
## PR description:

During testing, I noticed that the retrieve() method does not raise an error when an empty dataset is returned—for example, when a non-existent variable name is requested.

 - [x] Changelog is updated.


silviacaprioli:
Since this is breaking streaming functionality (the streaming goes on until dataset is empty), we agreed to put the raise in the retrieve method of the core diagnostic

oloapinivad (approved):

PR #2000, opened 2025-06-09
Custom diagnostic_name to use and combine different tool of diagnostics

### LINKED ISSUE #1857: Introduce a `diagnostic_name` for the diagnostic configuration file
### Feature request

An useful feature would be to include a `diagnostic_name` key in the dictionary of the configuration file of each diagnostic so that it would be possible to later change the name of the diagnostic currently hardcoded in the call of the outputsaver. This would allow for more correct grouping of the output when plotted into the dashboard and rationalise the organization of the files.

mnurisso:
## PR description:

As described in the issue we include in the configuration file the possibility to have a diagnostic_name in the configuration file. This allow for on the fly rename of the diagnostic in the filename.
This is used by diagnostics like radiation, which computes timeseries, biases etc that should be grouped in the radiation diagnostic.

Done:

- included the option in timeseries (timeseries, seasonal cycle and gregory CLI)
- gregory can now plot monthly only if 3 years of data are not available
- gregory is done by default by the radiation diagnostic
- timeseries and seasonal cycles are under the atmosphere or ocean grouping 

Close #1857 

 - [x] Changelog is updated.


mnurisso:
The PR is ready.
After a painful merge with the main now the new `Outputsaver` is in place together with the `diagnostic_name` customized.
The documentation on the config file is created and aligned with the PR.
The timeseries have now atmosphere and ocean instead of timeseries as diagnostic_name so that in a short future we can group them with the biases so that it is possible to focus on the realm while exploring the dashboard.
The gregory plot is now handled by the radiation diagnostic and the relative timeseries have now the radiation as `diagnostic_name`.

Ready for a review

config/diagnostics/timeseries/config_timeseries_atm.yaml
```diff
@@ -32,6 +32,7 @@ output:
 diagnostics:
   timeseries:
     run: true
+    diagnostic_name: 'atmosphere'
```
oloapinivad, line 35: uhm, atmosphere is not a diag name, at least not at this stage. what do you mean?

oloapinivad (approved):
It seems that most of the changes are technical updates and refinements. I am mostly okay with this, just a minor comments on the naming scheme used.

config/diagnostics/timeseries/config_timeseries_atm.yaml
mnurisso: I mean that with this PR the files will be saved as `atmosphere.timeseries` and `atmosphere.seasonalcycles`.
This was the motivation (together with radiation) for the code change. You don't want something similar?

PR #2014, opened 2025-06-19
Improving the eccodes shortname guessing using WMO tables 

oloapinivad:
As per title, I have been told that is possible to force the eccodes guessing toward WMO tables and avoid ECMWF tables. This can be done with `codes_set(gid, "centre", 0)`. I think that in the same we can explore ECMWF tables if needed

----

 - [ ] Changelog is updated.


oloapinivad:
This further improves the eccodes guessing by exploiting tables. It would prefer WMO GRIB2 over ECMWF tables as done in the older code. Ready for review for me

PR #2041, opened 2025-07-04
Realization support in `aqua-analysis`

mnurisso:
## PR description:

The PR allow to set `--realization` when calling the `aqua-analysis`.
The default output path is now with an additional layer which is:
- `r1` by default (no realization given)
- `r<number>` if the realization number is an int
- `<realization>` otherwise

The `--realization` flag is correctly passed to all the diagnostics (and `setup_checker`).
All the state-of-the-art diagnostics take a `--realization` flag, even fake, so that there is no failure.

For timeseries, teleconnections, setup_checker and global biases the `--realization` flag works also in the CLI. However no filename or other details are modified yet in this PR. This would require more careful discussion and feedback.

## People involved:

cc @jhardenberg @oloapinivad 

 - [x] Changelog is updated.


mnurisso:
Merged with the main, testing with a LRA from the historical ensemble, checking if the diagnostics are correctly called with the realization parameter

cli/aqua-analysis/aqua-analysis.py
```diff
@@ -125,6 +127,9 @@ def run_diagnostic_func(diagnostic: str, parallel: bool = False, regrid: str = N
     if catalog:
         extra_args += f" --catalog {catalog}"
 
+    if realization:
```
oloapinivad, line 130: why having as an extra arg instead of a default arg if now it is required by default in the LRA?

mnurisso: Beause we're not only working with LRA and if I have a source not LRA to analyze with the CLI I don't want to give an unused kwarg to the reader

src/aqua_diagnostics/ecmean/cli_ecmean.py
```diff
@@ -36,6 +36,8 @@ def parse_arguments(args):
                         help='source to be analysed')
     parser.add_argument('--regrid', type=str,
                         help='regrid method to be used [default: r100]', default='r100')
+    parser.add_argument('--realization', type=str, default=None,
```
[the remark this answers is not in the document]
mnurisso: It's minimal, I can take care of the conflicts

src/aqua/util/realizations.py
```diff
@@ -0,0 +1,22 @@
...
+    Args:
+        realization (str, optional): the realization string
+    
+    Returns:
+        str: the realization formatted to be used in folder or filename creation.
+    """
+    if not realization:
+        return "r1"
+    try:
+        # Try converting to int
```
oloapinivad, line 17: it looks cool but can't you use isdigit for this?

Issue #2050, opened 2025-07-07
[Tropical Rainfall] - diagnostic refactoring


CostanzaD:
## Tropical Rainfall

This issue coordinates the ongoing efforts to refactor and extend the tropical rainfall diagnostics:

Tropical precipitation diagnostics are currently running in low-resolution mode using a precomputed cache. No access to MSWEP on Lumi/MN5 so that it is unable to recompute.
Consider interoperability between high- and low-resolution workflows.

#### Tasks

- [ ] Devlop a `timsum` function in the `Reader` to support temporal aggregation
      ↳ @paolodavini

- [ ] Extract the PDF computation core function and move it somewhere as a basic function. Arguments as bins should be passed. 
       Placement (AQUA core or a new diagnostic module) can be decided later.
      ↳ @jost.hardenberg

- [ ] Refactor core function for LRA usage: allow passing `Reader` and `data` as arguments to support external context (e.g. lowres/highres interoperability)  
      ↳ @matteonurisso

- [ ] Isolate the graphical plotting function from the diagnostic and refactor as a reusable utility  
      ↳ @marco.cadau

#### Notes

- the usage of the stat field could be based as a ufunc or a string. if a string, browse the available function into a `lra_functions.py`. If a function, try to apply  the function to the data. The reader should be passed as an argument. 
- For tropical rainfall, the core function could be used from both the state-of-the-art version of the diagnostic and the frontier one.

PR #2061, opened 2025-07-08
Histogram as separate core function

### LINKED ISSUE #2055: Extract the PDF computation core function and move it somewhere as a basic function. Arguments as bins should be passed.

jhardenberg:
## PR description:

As discussed in #2055, part of the tasks to create a new treopical rainfall diagnostic based on LRA (#2050), this introduces a new 'histogram' function which accepts a dataarray and returns a dataarray with a histogram (i.e. the number of counts per class).
The arguments have been chosen to be as similar as possible to the classing numpy histogram function. It returns a DataArray containing the histogram. Latitudinal weighting is supported and on by default (this means that the 'counts' willl actually not be integers). If the input is a dask array it returns a lazy dask array using the dask histogram function and I would recommend to `.load()` it before plotting. If not it defaults to using the numpy histogram function.
A 'density' flag is supported to return a PDF instead of a histogram (intended for direct use).

Close #2055

----

 - [x] Changelog is updated.


oloapinivad:
Thanks! A comment about the placing, you put into the `core` of the diagnostics, I think I was suggesting to have it as a `core` function of the `Reader`. But of course we can discuss....

mnurisso:
I agree, the core of the diagnostics is now used for shared function that all the diagnostics need, so it doesn't fit particularly this definition

jhardenberg:
Hmmm... Isn't a histogram a core function that different diagnostics may need? 
I do not see why it should be a core function of the reader. Seems more logical as diagnostic core function. 
You would add it as a METHOD of the reader? It seems to specialized to me for that 
I added some (AI generated but nice and verified) tests

oloapinivad:
> ok, moving it. One problem: since I need to have the ability to change units, I import `from aqua.diagnostics.core import convert_data_units` .... but an import like this does not make sense in the framework (the framework should not depend on the diagnostics part)

well, `convert data units` should be part of the fixer probably no?
One suggestion where to place the histogram: I would put into the `timstat` module for now as independent method, not exposed through the reader for now. in the end, we compute histogram along the time dimension, no?

jhardenberg:
Lol, you guys let me keep moving this ....  
timstat is a class, while this one for now does not need to be a class.... 
So maybe leaving it in aqua.util is ok ?

jhardenberg:
Added docs, changelog. Running a test.

jhardenberg:
There are two things which could in theory be added to this, maybe it is worthwile:
1) add histogram as a method of the reader
2) (needed for 1) currently it works only with DataArrays, allow it to work with Datasets (looping over all variables)

jhardenberg:
There have been a few additions since your review @mnurisso , still ok with merge now?

PR #2078, opened 2025-07-14
New AQUA monthly fix

oloapinivad:
## PR description:
As per title


mnurisso (approved):

mnurisso:
> Ok, now monthly tests are actually working. Wondering if we want to move the matrix test to weekly frequency

I'd move it to weekly frequency when on lumi runners personally, but I'm not against increasing the frequency

PR #2107, opened 2025-07-29
Add norm keyword argument for customizable colorbar normalization

tovaz92:
This PR adds a new `norm` keyword argument to the `plot_single_map` function to enable more flexible and customizable control over color normalization when plotting maps.  
This is useful when using non-linear normalization (e.g., `BoundaryNorm`).

----

 - [x] Changelog is updated.


src/aqua/graphics/single_map.py
```diff
@@ -27,11 +27,12 @@ def plot_single_map(data: xr.DataArray,
                     extent: Optional[list] = None, coastlines: bool = True,
                     style: Optional[str] = None, figsize: tuple = (11, 8.5), nlevels: int = 11,
                     vmin: Optional[float] = None, vmax: Optional[float] = None, cmap: str = 'RdBu_r',
-                    cbar: bool = True, cbar_label: Optional[str] = None,
+                    cbar: bool = True, cbar_label: Optional[str] = None, 
+                    norm: Optional["matplotlib.colors.Normalize"] = None
```
mnurisso, line 31: I don't think it works with the " around the type.

src/aqua/graphics/single_map.py
```diff
@@ -137,18 +138,24 @@ def plot_single_map(data: xr.DataArray,
...
+            cs = data.plot.pcolormesh(ax=ax,
+                                      transform=ccrs.PlateCarree(),
+                                      cmap=cmap, norm=norm,
+                                      vmin=vmin, vmax=vmax,
+                                      add_colorbar=False)
+        else:
+            cs = data.plot.pcolormesh(ax=ax,
+                                      transform=ccrs.PlateCarree(),
+                                      cmap=cmap, norm=norm,
+                                      add_colorbar=False)
```
mnurisso, line 158: Why you needed to duplicate the pcolormesh call here?

mnurisso (changes requested):
I don't understand the duplication of the pcolormesh call.
Additionally please adapt the tests, there are many single_map and single_map_diff call that can be used to call once the function with norm, so that we test the functionality and we keep high coverage.

src/aqua/graphics/single_map.py
tovaz92: Ok, I suggest to put norm: Optional[object] instead ?

src/aqua/graphics/single_map.py
tovaz92: If norm is `not None` we need to remove the `vmin` and `vmax` args, otherwise an error arise as mpl has a conflicting vmax/vmin values.

To reduce repetition, it might be useful to define a common_kwargs_plot dictionary to store shared keyword arguments used across the four cases. What do you think?

src/aqua/graphics/single_map.py
tovaz92: Yes but this I suspect would imply importing the `colors` module just to add that specification line, I proposed to be a bit more generic to avoid this since the use of `norm` is quite common and not ambiguous. 
Anyway it is fine also to add the extra line if it's more explicit for readers.

oloapinivad: I see your point, please go aheda with the object solution

mnurisso (approved):

Issue #2124, opened 2025-08-11
Test new seaice and remove old


mnurisso:
We need to:

- [ ] test the new seaice in the aqua analysis and check if it correctly replace all the features of the old diagnostic
- [ ] remove the old code

@jhardenberg I'll finalize #2065 before this check

mnurisso:
#2065 is merged, time to check this

jhardenberg:
I did not see this issue on time so I opened #2132
There is aunit conversion issue when twsting the new seaice

mnurisso:
> I fixed the second issue in [#2140](https://github.com/DestinE-Climate-DT/AQUA/pull/2140) while as for the first one I do not know

I have an answer for the first one.
The ConfigStyle class has to be called before the grafical plot. This means that in the second plot the style is lost and go back to the default one. ConfigStyle has to be called again before the second plot

jhardenberg:
but `ConfigStyles` is called by `graphics/timeseries/plot_timeseries()` at the beginning and `plot_seaice.py` is just using  it .... ??? @mnurisso

jhardenberg:
I still do not understand..you do a series of `plot_timeseries()`and each of then. calls again `ConfigStyles` ....

tovaz92:
Update *on 1.*:
I see that if I initialize and call `ConfigStyle(style='aqua', loglevel=loglevel)` in the `PlotSeaIce` class before actually calling the plots, I no longer have the problem.

A possible reason could be that in `regions_type_plotter` the following is called first:
`fig, axes = plt.subplots(nrows=self.num_regions, ncols=1, figsize=(fig_height, 4 * self.num_regions), squeeze=False)` then afterwards `plot_timeseries` is called, which in turn calls `ConfigStyle`. 
The point could be that when I initialize the figure with subplots before calling `plot_timeseries`, I am anticipating the redefinition of the graphic style, which only happens afterwards, and perhaps this causes problems.

I would be inclined to think, as @jhardenberg  mentioned, that when I call `plot_timeseries` everything should still be overwritten, but this does not seem to happen.

I propose to simply add the call of the `ConfigStyle` when initialising the `PlotSeaIce` then and everything should work ok (and this should also fix the similar issue for the seasonal cycle too).

PR #2150, opened 2025-08-19
Initialise ConfigStyle to avoid override of the ConfigStyle parameters

tovaz92:
This mini PR fix the the Issue *1.* raised at [#2124](https://github.com/DestinE-Climate-DT/AQUA/issues/2124#issuecomment-3198282312)

The fix consist in calling `ConfigStyle` before any matplotlib object definition is done.

### DIFF
--- a/src/aqua_diagnostics/seaice/plot_seaice.py
+++ b/src/aqua_diagnostics/seaice/plot_seaice.py
@@ -10,6 +10,7 @@ from aqua.util import ConfigPath
 from aqua.graphics import plot_timeseries, plot_seasonalcycle
 from collections import defaultdict
 from .util import defaultdict_to_dict, extract_dates, _check_list_regions_type
+from aqua.graphics import ConfigStyle
 
 xr.set_options(keep_attrs=True)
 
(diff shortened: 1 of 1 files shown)


src/aqua_diagnostics/seaice/plot_seaice.py
```diff
@@ -64,6 +65,8 @@ def __init__(self, monthly_models=None, annual_models=None,
             loglevel (str, optional): 
                 Logging level for debugging and information messages. Defaults to 'WARNING'.
         """
+        ConfigStyle(style='aqua', loglevel=loglevel)
```
jhardenberg, line 68: This hardcodes the style, while in genral we would like the style to be defined from the aqua config file

jhardenberg: also, I am not so sure that it is good to have this in the class init...maybe better in `plot_seaice`

jhardenberg: just call without the style keyword and it will default to the style of the config file

src/aqua_diagnostics/timeseries/plot_timeseries.py
```diff
@@ -13,7 +13,7 @@ def __init__(self, diagnostic_name: str = 'timeseries',
                  ref_monthly_data=None, ref_annual_data=None,
                  std_hourly_data=None, std_daily_data=None,
                  std_monthly_data=None, std_annual_data=None,
-                 loglevel: str = 'WARNING'):
+                 style=None, loglevel: str = 'WARNING'):
```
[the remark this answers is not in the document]
tovaz92: removed now

mnurisso (approved):

PR #2157, opened 2025-08-20
[Gregory plot] Small change to avoid try-except in set_label

tovaz92:
Follow the #TODO comment to avoid Try/Except with a small refactoring.


jhardenberg:
I would add the diagnostic this refers to in the title

mnurisso (approved):

mnurisso:
Adding changelog so that tests are triggered, then we can merge

PR #2190, opened 2025-08-27
Split out plotting function for vertical profile

### LINKED ISSUE #2034: GlobalMean VerticalBias part of core graphics function
### Feature request

As per title, and add also the possibility to plot the full reference field as a contour

silviacaprioli:
## PR description:

Extract vertical profile plotting function from GlobalBiases diagnostic and move it to aqua graphics module

Close #2034

## People involved:

----

 - [ ] Changelog is updated.

### DIFF
--- a/src/aqua/graphics/vertical_profiles.py
+++ b/src/aqua/graphics/vertical_profiles.py
@@ -1,0 +1,72 @@
+from typing import Optional
+import matplotlib.pyplot as plt
+import numpy as np
+import xarray as xr
+from aqua.logger import log_configure
+from .styles import ConfigStyle
+
+def plot_vertical_profile(data: xr.DataArray, var: str,
+                         plev_min: Optional[float] = None, plev_max: Optional[float] = None, 
+                         vmin: Optional[float] = None, vmax: Optional[float] = None, 
+                         nlevels: Optional[int] = 18,  
+                         title: Optional[str] = None, style: Optional[str] = None,
+                         return_fig: bool = False, 
+                         loglevel='WARNING',  **kwargs):
+    """
+    Plots a zonal mean vertical profile 
+
+    Args:
+        data (xarray.DataArray): data to plot.
+        var (str): Variable name to plot (for labeling purposes).
+        vmin (float, optional): Minimum colorbar value.
+        vmax (float, optional): Maximum colorbar value.
+        plev_min (float, optional): Minimum pressure level to plot.
+        plev_max (float, optional): Maximum pressure level to plot.
+        nlevels (int, optional): Number of contour levels for the plot.
+        title (str, optional): Title of the plot.
+        return_fig (bool, optional): If True, returns the figure and axis objects. Defaults to False.
+        style (str, optional): Style to use. Defaults to None (aqua style).
+    """
+
+    logger = log_configure(loglevel, 'plot_single_map')
+    ConfigStyle(style=style, loglevel=loglevel)
+
+    if plev_min is None:
+        plev_min = data['plev'].min().item()
+    if plev_max is None:
+        plev_max = data['plev'].max().item()
+    # Slice pressure levels 
+    mask = (data['plev'] >= plev_min) & (data['plev'] <= plev_max)
+    data = data.sel(plev=data['plev'].where(mask, drop=True))
+    # Ensure reasonable number of levels
+    nlevels = max(2, int(nlevels))
+    zonal_mean = data.mean(dim='lon')
+    
+    # Determine colorbar limits if not provided
+    if vmin is None or vmax is None:
+        vmin, vmax = float(zonal_mean.min()), float(zonal_mean.max())
+        if vmin * vmax < 0:
+            vmax = max(abs(vmin), abs(vmax))
+            vmin = -vmax
+
+    levels = np.linspace(vmin, vmax, nlevels)
+
+    fig, ax = plt.subplots(figsize=(10, 8))
+    cax = ax.contourf(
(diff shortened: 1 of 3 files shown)


silviacaprioli:
Yes I know, I'm on it. I would add it here, not in another PR. 
I opened the PR but it's not ready to merge

silviacaprioli:
image

src/aqua/graphics/vertical_profiles.py
```diff
@@ -0,0 +1,168 @@
...
+    if plev_min is None:
+        plev_min = data['plev'].min().item()
+    if plev_max is None:
+        plev_max = data['plev'].max().item()
+    # Slice pressure levels 
+    mask = (data['plev'] >= plev_min) & (data['plev'] <= plev_max)
+    data = data.sel(plev=data['plev'].where(mask, drop=True))
+    # Ensure reasonable number of levels
+    nlevels = max(2, int(nlevels))
+    zonal_mean = data.mean(dim='lon')
```
mnurisso, line 44: shouldn't the mean be done outside of the plotting function?

src/aqua/graphics/vertical_profiles.py
```diff
@@ -0,0 +1,168 @@
...
+        nlevels (int, optional): Number of contour levels for the plot.
+        title (str, optional): Title of the plot.
+        return_fig (bool, optional): If True, returns the figure and axis objects. Defaults to False.
+        style (str, optional): Style to use. Defaults to None (aqua style).
+    """
+
+    logger = log_configure(loglevel, 'plot_vertical_profile')
+    ConfigStyle(style=style, loglevel=loglevel)
+
+    if plev_min is None:
```
mnurisso, line 35: We are forcing this to be only an atmospheric function, while this could be used by ocean as well, do you think you can have a `lev_name` argument which is `plev` by default but can be adapted so that you keep the flexibility?

src/aqua/graphics/vertical_profiles.py
```diff
@@ -0,0 +1,168 @@
...
+    # Determine colorbar limits if not provided
+    if vmin is None or vmax is None:
+        vmin, vmax = float(zonal_mean.min()), float(zonal_mean.max())
+        if vmin * vmax < 0:
+            vmax = max(abs(vmin), abs(vmax))
+            vmin = -vmax
+
+    levels = np.linspace(vmin, vmax, nlevels)
+
+    fig, ax = plt.subplots(figsize=(10, 8))
```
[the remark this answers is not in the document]
mnurisso: I'd also do, as in the other functions, the option to give fig and ax as an input, I think this function can be used by @ghosh97 as well in the trend

src/aqua/graphics/vertical_profiles.py
```diff
@@ -0,0 +1,168 @@
...
+    cax = ax.contourf(
+        zonal_mean['lat'], zonal_mean['plev'], zonal_mean,
+        cmap='RdBu_r', levels=levels, extend='both'
+    )
+    ax.set_yscale('log')
+    ax.set_ylabel('Pressure Level (Pa)')
+    ax.set_xlabel('Latitude')
+    ax.invert_yaxis()
```
mnurisso, line 63: This only if the name is `plev`

src/aqua/graphics/vertical_profiles.py
```diff
@@ -0,0 +1,168 @@
...
+    if title:
+        logger.debug("Setting title to %s", title)
+        ax.set_title(title)
+
+    if return_fig:
+        logger.debug("Returning figure and axes")
+        return fig, ax
+
+
+def plot_vertical_profile_diff(data: xr.DataArray, data_ref: xr.DataArray,
```
mnurisso, line 76: This is not using internally `plot_vertical_profile`, could it be a solution similar to what is done in the maps?

mnurisso (commented):
I think with little effort we can do a good generalization of the function you wrote.
Also, unfortunately I would require test and maybe a documentation update before merging this.

src/aqua/graphics/vertical_profiles.py
```diff
@@ -0,0 +1,168 @@
...
+        data_ref (xr.DataArray): Reference data to plot the difference.
+        var (str): Variable name to plot (for labeling purposes).
+        vmin (float, optional): Minimum colorbar value.
+        vmax (float, optional): Maximum colorbar value.
+        plev_min (float, optional): Minimum pressure level to plot.
+        plev_max (float, optional): Maximum pressure level to plot.
+        nlevels (int, optional): Number of contour levels for the plot.
+        title (str, optional): Title of the plot.
+        return_fig (bool, optional): If True, returns the figure and axis objects. Defaults to False.
+        style (str, optional): Style to use. Defaults to None (aqua style).
```
mnurisso, line 101: style is not used here, isn't it?

src/aqua/graphics/vertical_profiles.py
silviacaprioli: mmh I don't know, I think i would like the function to be able to plot the vertical profile by just giving it the original data array. Are there cases in which we don't want the zonal mean?

mnurisso: I think in principle no operation on data should be done inside the plotting function. For example as a user I'd have used as input data something only with lat and plev, generating an error because of this line of code. I have no clear idea of a case where we should plot meridional instead of zonal, but again I don't see why forcing it here

silviacaprioli:
I've generalized also the x coord (lat by default). Added tests, a notebook and updated the docs. I think this should be ready to merge

PR #2193, opened 2025-08-28
Timeseries description update with longname

### LINKED ISSUE #2154: Add long variable name to legend of timeseries
### Feature request

The description of Timeseries and SeasonalCycles currently does not mention the variable, fore example:
_"Description: Time series for climatedt-phase1 IFS-NEMO ssp370 with reference ERA5 with standard deviation from 19900101 to 20201231"_
and the short variable name is only in the figure title ù8whic is part of the image). This prevents from rapidly searching for the figure of interest in a dashboard. I would include the long variable name, as currently done for global biases, like
_"Description: Time series of of 2 metre temperature for climatedt-phase1 IFS-NEMO ssp370 with reference ERA5 with standard deviation from 19900101 to 20201231"_

mnurisso:
## PR description:

Move to the attributes deduction everything which is feasible in the timeseries and seasonalcycle diagnostics.
Update of the description as requested in #2154 

Close #2154 

 - [ ] Changelog is updated.

### DIFF
--- a/src/aqua_diagnostics/timeseries/base.py
+++ b/src/aqua_diagnostics/timeseries/base.py
@@ -331,28 +331,25 @@ class PlotBaseMixin():
         self.logger.debug('Reference label: %s', ref_label)
         return ref_label
 
-    def set_title(self, region: str = None, var: str = None, units: str = None, diagnostic: str = None):
+    def set_title(self, diagnostic: str = None):
         """
-        Set the title for the plot.
+        Set the title for the plot. Uses standard_name, long_name, and units attributes.
 
         Args:
-            region (str): Region to be used in the title.
-            var (str): Variable name to be used in the title.
-            units (str): Units of the variable to be used in the title.
             diagnostic (str): Diagnostic name to be used in the title.
 
         Returns:
             title (str): Title for the plot.
         """
         title = f'{diagnostic} '
-        if var is not None:
-            title += f'for {var} '
+        if self.standard_name is not None:
+            title += f'for {self.standard_name} '
 
-        if units is not None:
-            title += f'[{units}] '
+        if self.units is not None:
+            title += f'[{self.units}] '
 
-        if region is not None:
-            title += f'[{region}] '
+        if self.region is not None:
+            title += f'[{self.region}] '
 
         if self.len_data == 1:
             title += f'for {self.catalogs[0]} {self.models[0]} {self.exps[0]} '
--- a/src/aqua_diagnostics/timeseries/base.py
+++ b/src/aqua_diagnostics/timeseries/base.py
@@ -279,6 +279,7 @@ class PlotBaseMixin():
         self.diagnostic_name = diagnostic_name
         log_name = 'Plot' + diagnostic_name.capitalize()
         self.logger = log_configure(log_level=loglevel, log_name=log_name)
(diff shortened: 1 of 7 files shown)


mnurisso:
Example of description:

'Time series of 2 metre temperature [degC] for climatedt-phase1 ICON historical-1990 with reference ERA5 with standard deviation from 1990-01-01 to 1999-12-31. The shaded area represents 2 standard deviations.'

src/aqua_diagnostics/timeseries/base.py
```diff
@@ -2,8 +2,7 @@
 import xarray as xr
 from aqua.fixer import EvaluateFormula
 from aqua.logger import log_configure
-from aqua.util import ConfigPath
-from aqua.util import frequency_string_to_pandas, time_to_string
+from aqua.util import frequency_string_to_pandas, time_to_string, strlist_to_phrase
```
jhardenberg, line 5: As discussed also in another PR, would these utilities (used only by diagnostics) be better in the diagnostic core functions?

src/aqua_diagnostics/timeseries/base.py
```diff
@@ -420,10 +427,10 @@ def save_plot(self, fig, var: str = None, description: str = None, region: str =
         metadata = {"Description": description, "dpi": dpi }
         extra_keys = {'diagnostic_product': diagnostic_product}
 
-        if var is not None:
-            extra_keys.update({'var': var})
-        if region is not None:
-            region = region.replace(' ', '').lower()
+        if self.standard_name is not None:
+            extra_keys.update({'var': self.standard_name})
```
jhardenberg, line 431: This creates filenames including standard_name instead of var. Actually I believe this should be discussed.... I sort of like and think it is pretty compact dealing filenames just with the short name, no?

jhardenberg: If self.standard name is None (I do not see how, since you fix it earlier) no var info is include din the filename?

jhardenberg: I did not yet give approval because I believe we should settle this

src/aqua_diagnostics/timeseries/base.py
mnurisso: We can discuss it. As probably I said there already I still prefer to have these utilities in the aqua framework. I don't see for example why something like `time_to_string` couldn't be useful to update the `history` attribute in the `Reader`

mnurisso:
> at line 195 of plot_timeseries.py you still have `var (str): Variable name to be used in the title and description.`

I left also region, I'll remove both

mnurisso:
I've made the required changes, let me know if it looks ok now

config/diagnostics/timeseries/config_timeseries_atm.yaml
```diff
@@ -55,10 +55,10 @@ diagnostics:
         std_startdate: '19900101'
         std_enddate: '20201231'
       tnlwrf+tnswrf:
-        standard_name: "net_top_radiation"
+        short_name: "net_top_radiation"
```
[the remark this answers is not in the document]
jhardenberg: Ah, ok. I still wonder if it would not be better to add a new variable to describe netTOA in the fixer but indeed it would not have agrib code...a discussion for another time

PR #2218, opened 2025-09-09
Revert `xarray_to_pandas_freq`

### LINKED ISSUE #2215: Time frequency estimation introduced in Diagnostic can fail
### machine

Lumi

### catalog

Others (specify)

### version

main

### What happened?

The check for having enough months introduced in the `Diagnostic` class in #2205 does not work in all cases (it fails with EC-Earth data for example). In fact the pandas funtion `pd.infer_freq()` does not work and returns `Null`, presumably because the monthly dates look like this:
````
array(['1990-01-16T12:00:00.000000000', '1990-02-15T00:00:00.000000000',
       '1990-03-16T12:00:00.000000000', '1990-04-16T00:00:00.000000000',
       '1990-05-16T12:00:00.000000000', '1990-06-16T00:00:00.000000000',
````
sometimes on the 15th, sometimes on the 16th, sometimes at 00:00 sometimes at 12:00.....

I left a sample of data to test on lumi to test

```python
import xarray as xr
from aqua.util import xarray_to_pandas_freq
data = xr.open_dataset("/scratch/project_465000454/jvonhar/msl.nc")
print(xarray_to_pandas_freq(data))
None
```

### Error messages

_No response_

None

mnurisso:
## PR description:

Revert function modification since CMOR monthly data are not compatible with pandas `infer_freq` function

Close #2215

## People involved:

cc @jhardenberg 

 - [ ] Changelog is updated.

### DIFF
--- a/src/aqua/util/time.py
+++ b/src/aqua/util/time.py
@@ -1,6 +1,7 @@
 """
 Module including time utilities for AQUA
 """
+import math
 import numpy as np
 import pandas as pd
 import xarray as xr
@@ -52,9 +53,8 @@ def frequency_string_to_pandas(freq):
 
 def xarray_to_pandas_freq(xdataset: xr.Dataset | xr.DataArray):
     """
-    Given a Xarray Dataset or DataArray, estimate the time frequency
-    with the pandas method.
-    Important: the infer_freq method requires at least 3 time points.
+    Given a Xarray Dataset, estimate the time frequency and convert
+    it as a Pandas frequency string
 
     Args:
         xdataset (xr.Dataset | xr.DataArray): The input xarray object.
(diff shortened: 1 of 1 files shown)


src/aqua/util/time.py
```diff
@@ -98,12 +111,14 @@ def pandas_freq_to_string(freq: str) -> str:
         'Q-NOV': 'seasonal',
         'Q': 'seasonal',
         # Monthly
+        '1MS': 'monthly',
```
jhardenberg, line 114: Could it happen that this is 2MS ? Should we ignore leading digits?

jhardenberg (commented):
tested and all works

jhardenberg (approved):
ok

Issue #2234, opened 2025-09-11
Rename LRA to DROP (Data Reduction Operator)


oloapinivad:
### Feature request

The extension of the LRA tool requires a renaming to avoid disambiguations. The idea is that the LRA will always be the monthly reduced archive, but the tool, currently named `LRAgenerator` will be renamed as `DROP`. 

This has to be implemented in the code and in the documentation. Clear distinction between what is the LRA must be preserved. 

- [x] Rename the `LRAgenerator` class
- [x] Folder names
- [x] Entry point
- [x] Parallel LRA tool
- [x]  Source name has to be adjusted: we will drop lra for all the instances but the default `lra-r100-monthly`
- [x] update documentation explaining DROP usage

PR #2248, opened 2025-09-15
Rename LRA to DROP (Data Reduction OPerator)

mcadau:
## PR description:
As discussed offline  and in #2234, despite the specific source `lra-r100-monthly` still being LRA (Low Resolution Archive), the LRA generator changes its name to DROP - Data Reduction OPerator. 

The catalog entry sources other than `lra-r100-monthly` will no longer present the `lra` prefix.

#2234 

----

 - [x] Changelog is updated.


mcadau:
Yep sure! I will specify this in the documentation and where required in the code

mcadau:
Okay now I should have updated everything, but I am afraid something may be still missing. The current tests seem to work, but, as discussed yesterday, I haven't changed the structure (only the naming) of the DROP notebook which maybe will be more adequate to show that everything is fine right now. I anyway set it as ready for review and I'll intervene in case of further modification needed. In addition, I'll open the issue about the DROP notebook not working

src/aqua/drop/drop.py
```diff
@@ -21,7 +21,7 @@
 from aqua.util import ConfigPath, file_is_complete
 from aqua.util import create_zarr_reference
 from aqua.util import area_selection, replace_intake_vars
-from .lra_util import move_tmp_files, list_lra_files_complete
+from .drop_util import move_tmp_files, list_lra_files_complete
```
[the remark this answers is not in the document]
oloapinivad: nope, I think that the tool checks the consistency of all the files produced by the former class LRAgenerator, so I think that renaming would be ok. but please double check the code flow...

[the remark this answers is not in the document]
mcadau: Ok perfect!

oloapinivad:
I think that the technical implementation is the same so that the documentation can mostly go with a simple name change. What would be nice to add is explain well the difference between the original idea. no longer a tool to produce the LRA only, but a tool to extract data from whatever dataset. you could also keep a section on the LRA, to explain what is actually is in this context

mnurisso:
I think the documentation has to be checked and updated carefully with the new meaning, not only a name change.
See [LRA docs](https://aqua.readthedocs.io/en/latest/lra.html), there is a section explaining what LRA is. This is fine, but now there should be something explaining the DROP philosophy and the what specific product LRA is.

src/aqua/drop/catalog_entry_builder.py
```diff
@@ -53,10 +53,14 @@ def create_entry_name(self):
         """
         Create an entry name for the LRA
         """
-
-        entry_name = f'lra-{self.resolution}-{self.frequency}'
+        # Default LRA entry keeps the 'lra-' prefix
+        if self.resolution == 'r100' and self.frequency == 'monthly':
+            entry_name = f'lra-{self.resolution}-{self.frequency}'
+        else:
+            # All other entries drop the 'lra-' prefix
```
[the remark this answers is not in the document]
oloapinivad: yep the lower case `lra` as prefix of `r100-monthly` case.

src/aqua/cli/parser.py
```diff
@@ -7,7 +7,7 @@
 from aqua import __version__ as version
 from aqua import __path__ as pypath
 from aqua.cli.analysis import analysis_parser
-from aqua.cli.lra import lra_parser
+from AQUA.src.aqua.cli.drop import drop_parser
```
mnurisso, line 10: typo

src/aqua/drop/catalog_entry_builder.py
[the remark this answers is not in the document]
mcadau: I got back to this version: as discussed also with @mnurisso, lra will be present only in lra-100-monthly, and nor lra or drop will be present for the rest of the catalog entries. 
If we change our mind, we can discuss it

oloapinivad:
yep, I would keep the same philisophy also for the zarr part.

mcadau:
I guess this PR is now ready for a second review. 

I highlight that it's highly likely that the `/pfs/lustrep3/appl/local/climatedt/data/AQUA/LRA` folder on LUMI has to be renamed to DROP (I wasn't allowed) since a few files I renamed utilise this folder.
The renaming is now much more complete, basically keeping LRA only when referring to the specific `lra-r100-monthly` entry, and DROP more or less everywhere else in the AQUA repository. Documentation has also been updated accordingly. 

Maybe an analogous PR will be necessary for the Climate-DT-catalog repo?

oloapinivad:
> I highlight that it's highly likely that the `/pfs/lustrep3/appl/local/climatedt/data/AQUA/LRA` folder on LUMI has to be renamed to DROP (I wasn't allowed) since a few files I renamed utilise this folder. 

> Maybe an analogous PR will be necessary for the Climate-DT-catalog repo?

I am not sure we should do this, in the end the LRA is object DROP produces in the speciifc . So we can leave it as it here for the LRA files produced in the past, correct?

cli/drop/icon_drop_surface_ocean.sh
```diff
@@ -24,8 +24,8 @@ exp="historical-1990"
 for variable in avg_sos avg_tos ; do
```
oloapinivad, line 24: revert all the changes in this file. This was a script specifically to fix ICON data, which in principle is completely deprecated, but will always need to point to LRA, not to DROP

docs/sphinx/source/stateoftheart_diagnostics/ecmean.rst
```diff
@@ -84,21 +84,25 @@ For **Global Means**, the following variables are requested
 For both diagnostics, if a variable (or more) is missing, blank line will be reported in the output figures. 
 
 .. note ::
-    ECmean4 is made to work with CMOR variables, but can handle name and file conversion with specification of
-    an `interface file <https://ecmean4.readthedocs.io/en/latest/configuration.html#interface-files>`_.
+    ECmean4 is made to work with CMOR variables, but can handle name and file conversion with 
+    specification of an `interface file <https://ecmean4.readthedocs.io/en/latest/configuration.html#interface-files>`_.
```
oloapinivad, line 88: as for boxplot,  I do not understand what has changed here

docs/sphinx/source/drop.rst
```diff
@@ -0,0 +1,268 @@
+.. _drop:
+
+DROP - Data Reduction OPerator
+===============================================
+
+DROP (Data Reduction OPerator) is a comprehensive tool within the AQUA framework designed 
+to extract, process, and organize data from any dataset, with particular focus on climate 
+data reduction and archival.
```
oloapinivad, line 8: "with particular focus on climate data reduction and archival."not fully sure I understand, DROP works only for climate data.

docs/sphinx/source/drop.rst
```diff
@@ -0,0 +1,268 @@
...
+
+
+The Low Resolution Archive (LRA) Context
+----------------------------------------
+
+The Low Resolution Archive is a key use case for DROP. The LRA is an intermediate layer 
+of data reduction that simplifies analysis of extreme high-resolution data by providing:
+
+- Monthly data at coarse resolution
+- Fast access for climate model assessment
```
oloapinivad, line 62: these two points are not necessary. Please mention that the LRA case isa case where drop is set to monthly and r100, this would be enough no?

docs/sphinx/source/drop.rst
```diff
@@ -0,0 +1,268 @@
...
+===============================================
+
+DROP (Data Reduction OPerator) is a comprehensive tool within the AQUA framework designed 
+to extract, process, and organize data from any dataset, with particular focus on climate 
+data reduction and archival.
+
+
+What is DROP?
+-------------
+
```
oloapinivad, line 13: I would add a warning stating that from version 0.18 DROP replace LRA.

src/aqua/cli/drop.py
```diff
@@ -2,28 +2,28 @@
 # -*- coding: utf-8 -*-
 '''
 AQUA regridding tool to create low resolution archive.
-Make use of aqua.LRA_Generator class to perform the regridding.
+Make use of aqua.DROP class to perform the regridding.
```
oloapinivad, line 5: aqua.Drop

src/aqua/drop/drop_util.py
```diff
@@ -1,4 +1,4 @@
-"""Utility module for LRA/OPA"""
+"""Utility module for DROP/OPA"""
```
oloapinivad, line 1: I think you can remove opa from here

templates/drop/config_drop.tmpl
```diff
@@ -1,4 +1,4 @@
-# templates for LRA generation. Please modify everything between {{ }} according to your configuration
+# templates for DROP-generated outputs. Please modify everything between {{ }} according to your configuration
```
oloapinivad, line 1: I wondering if you could duplicate this file, by providing a `config_drop.tmpl` with some more advanced options, and keep the `config_lra.tmpl` for the standard LRA usage.

templates/drop/config_drop_workflow.tmpl
```diff
@@ -14,7 +14,7 @@ paths:
 options:  
   loglevel: INFO
 
-# all the variable listed here will be processed by the LRA creator
+# all the variable listed here will be processed by the DROP user
```
oloapinivad, line 17: this is - at this stage - still a LRA call in the workflow. please revert the comment

templates/drop/drop-submitter_container.tmpl
```diff
@@ -39,15 +39,15 @@ singularity exec   \
 $AQUA_container  \
 bash -c \
 '
-cd /users/caprioli/AQUA/cli/lra
-aqua lra --config ifs-nemo-test.yaml -w $WORKERS -d -l INFO
+cd /users/caprioli/AQUA/cli/drop
```
oloapinivad, line 42: for being a template this is a bit wrong. Can we have some more generic than silvia's path?

docs/sphinx/source/stateoftheart_diagnostics/boxplots.rst
```diff
@@ -21,17 +22,24 @@ There is one class for the analysis and one for the plotting:
 File structure
 --------------
```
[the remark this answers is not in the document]
oloapinivad: well, diagnostics are based on the LRA, not on the DROP output, correct? but I might misunderstood what you mean here...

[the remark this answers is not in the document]
mcadau: They are, indeed, in fact most of the pseudo-changes of the documentation is about just specifying where LRA comes from (+ linting)

docs/sphinx/source/drop.rst
```diff
@@ -0,0 +1,268 @@
...
+- Multiple statistics: mean, std, max, min
+- Handling of incomplete time chunks
+
+**Spatial Processing:**
+- Regridding to any supported resolution or native grid
+- Regional data extraction with configurable boundaries  
+- Support for both regular and irregular grids
+
+**Data Management:**
+- Automatic catalog entry generation
```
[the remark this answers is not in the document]
mcadau: Ah ok great I think I understood what you mean! So something like:

"- Automatic catalog entry generation for DROP-generated outputs"

would be enough?

PR #2262, opened 2025-09-17
[Analysis] Add new hovmoller

jhardenberg:
## PR description:

This runs the new refactored  Hovmöller diagnostic instead of ocean3d trends in aqua analysis.
The grouping  confighas been adapted to recognize the new output filenames.

----

 - [ ] Changelog is updated.


mnurisso:
> I actually realize now: is Ocean3D trends doing something else apart from the Hovmoller?

I think there are the 2d maps of trends ad different levels as well which will be together ([these ones](https://aqua-web-climatedt.2.rahtiapp.fi/exp_climatedt-phase1_IFS-NEMO_ssp370_r1/#trends))

PR #2368, opened 2025-10-22
introducing startdate enddate in aqua analysis

oloapinivad:
## PR description:

This PR brings in the possibility of specifyin in aqua analysis and in all the CLI a startdate and enddate from command line. 
To be refined yet. 

 - [ ] Changelog is updated.


mnurisso:
With this PR the argument is there but used only by ecmean, is it correct?

mnurisso:
> Yes diagnostics has to be modified in case they are not using. I would wait for merging of the centralized cli where we can make a bit of cleaning to extend this to all the diags, do you agree?

Since this is not needed in the operative cycle yes

Issue #2253, opened 2025-09-16
Diagnostics updated grouping and observational dataset


mnurisso:
This table summarizes how the plot will be shown and how the config files will be grouped in the final implementation for the new cycle.
 
Category | Sub-categories (actual diagnostics) | Observational dataset | Extra tasks | Status 
-- | -- | -- | -- | --
Radiation (TOA) | Timeseries, Seasonal cycle, Boxplot, Bias | CERES ebaf-toa42 | Split the “metadiag”, write configuration file, include cloud forcing from bias | 🟠 
Radiation (Surface) | Timeseries, Seasonal Cycle, Boxplot, Bias | CERES ebaf-sfc42 | Split the “metadiag”, write configuration file, add timeseries for net surface`*`| 🟠 
Sea Ice | Timeseries, Seasonal cycle, Bias | OSI-SAF (conc), PSC (vol) | Only update dataset| 🟢 
Water Cycle | Timeseries, Seasonal cycle, Diurnal cycle`**`, Bias, Lat/Lon, Tropical rainfall`**`, Global p-e | MSWEP | Create from scratch| 🟠 
Atmosphere 2D | Timeseries, Seasonal bias, Seasonal cycle, Bias | ERA5 | Clean the current configuration file| 🟢 
Atmosphere 3D | Bias on levels, Bias vertical | ERA5 | Clean the current configuration file| 🟢
Climate Metrics | PI, Global Mean,Gregory plot | internal for PI/GM, CERES+Berkley | Less urgent, see if they can be merged| 🟢
Ocean 2D | Timeseries, Seasonal cycle | ESA-CCI | Update dataset, focus on regions| 🟢
Ocean 3D | Hovmöller, Trend, Stratification, Mixed Layer Depth, Multi level Timeseries | EN4 for circulation | As it is| 🟢
Teleconnections | ENSO, NAO | ERA5 | Understand config and cli separation for ENSO and NAO | 🟠 

`*` Boxplot will require #2224
`**` Not available yet in this cycle

Sub-issue will follow for the implementation of the individual diagnostics

jhardenberg:
Why has tropical rainfall been erased from Water Cycle?

jhardenberg:
This can be very confusing... We should list here the desired ultimate grouping. We could place a * close to the diagnostics which we do not expect to be included in real 'metadiagnostics' in this cycle, but still this could document how we would like them to be ultimately organized

Issue #1910, opened 2025-05-13
Refactor `aqua_analysis` to provide more flexible combination of diagnostics


oloapinivad:
### Feature request

As per title, more asap

mnurisso:
We need to clarify more what has to be done here so that @jhardenberg and me we can start working on the task

jhardenberg:
what do you think? @mnurisso ?

jhardenberg:
Ok, I gave it more thought, so here is a draft of how the config file could really look like. 

@mnurisso @oloapinivad @silviacaprioli 
Please check it and let me know if you agree and then I can proceed to implement it modifying aqua-analysis accordingly.
The groupings are loosely following what we will implement in the dashboard but, while I would align them as much as possible, I would keep the two concepts separate. The ultimate way how to group figures will still be determined by `config.grouping.yaml `(which we need to move to config / .aqua anyway). This is as close as we can probably get with the current diagnostics. The 'diagnostic_name' of each of them is still specified in the config file.

The main features are:

- an initial collection of cli tools (diagnostics) which can be used to compose the following 'metadiagnostics', under `cli:`

- a list of the actual diagnostics to run (also organized in metadiagnostic/diagnostic). These could be set to true/false or commented. There are several groups (group1, group2) and the idea is that a diagnostic in a group will be run only if all diagnostic in the previous grup have finished.

- The metadiagnostics like
```
atmosphere:
    timeseries:
       ....
   global_biases:
    ....
```
each of them with separate config files.
We still have the `nworkers` key but remember that (it is already like this now) it only specifies how many workers to use if a global dask server ("cluster") is not used and instead each diagnostic has its own separate server. The default is to use the global cluster. If `nworkers` is not specified it defaults to the value specified in the `job:` section .

Ok, here is a sample config file for aqua-analysis:

````
 # Configuration file for aqua-analysis.sh

# Variables $AQUA, $AQUA_CONFIG and $OUTPUT can be used to refer to the root directory of AQUA,
# the config directory (i.e. .aqua/diagnostics) and the output directory respectively

job:
    # The max_threads variable serves as a mechanism to control the maximum number of diagnostics being run concurrently
    # - If max_threads is set to 0 or a negative value: There is no limit on the number of parallel diagnostics,
    #                                                   and all processes run in parallel without waiting. 
    #                                                   This is suitable for situations where you want to utilize
    #                                                   the maximum available resources without any restrictions.
[...]

PR #2371, opened 2025-10-23
Config file setup and aqua analysis updates

mnurisso:
## PR description:

- update of the config file to a final setup
- update of aqua config file to have metadiagnostics
- update of aqua analysis 

Close https://github.com/DestinE-Climate-DT/AQUA/issues/2253 close https://github.com/DestinE-Climate-DT/AQUA/issues/1910

## People involved:

----

 - [x] Changelog is updated.

### DIFF
--- a/src/aqua/analysis/analysis.py
+++ b/src/aqua/analysis/analysis.py
@@ -82,7 +82,7 @@ def _build_extra_args(**kwargs):
 
 
 def run_diagnostic_func(diagnostic: str, parallel: bool = False,
-                        regrid: str = None, cli: dict = None,
+                        regrid: str = None, cli: dict = {},
                         diag_config=None, catalog=None, model='default_model', exp='default_exp',
                         source='default_source', source_oce=None,
                         startdate=None, enddate=None, realization=None,
@@ -110,10 +110,6 @@ def run_diagnostic_func(diagnostic: str, parallel: bool = False,
         logger: Logger instance for logging messages.
         cluster: Dask cluster scheduler address.
     """
-
-    # safety check on clu
-    if cli is None:
-        cli = {}
 
     # Internal naming scheme:
     # diagnostic: the name of the wrapper metadiagnostic, e.g. atmosphere2d, climate_metrics, etc.


jhardenberg:
ecmean does not respect diagnostic name when generating figures:
image
@oloapinivad

jhardenberg:
For teleconnections the output  figure filenames do not have the diagnostic name (teleconnections) at the beginning. Should "nao" etc not go into diagnostic_product ? (the second key)
image
 @mnurisso

jhardenberg:
Ocean3d figures filenames have blanks in filenames (this is bad practice on unix filesystems)
image

Also: MLD does not respect the diagnostic_name in the output (see "MLD" as first key)

@ghosh97

silviacaprioli:
> Output netcdf files of the radiation diagnostic for the boxplots tool do not respect diagnostic name: 
> 
Fixed

jhardenberg:
> ecmean does not respect diagnostic name when generating figures: image @oloapinivad

Ok, this is now solved

ghosh97:
>  @ghosh97 please check the ocean stratification one. (fix in the main, since the fail is present also there, so please create a new PR once this is corrected) 

opened a PR for this: https://github.com/DestinE-Climate-DT/AQUA/pull/2378

jhardenberg:
>  Ocean3d figures filenames have blanks in filenames (this is bad practice on unix filesystems)

Now fixed in ocean_trends

jhardenberg:
I still have this for trends, experiment with other chunking options
image

jhardenberg:
To summarize the situation:
- stratification is still not working (see comment in #2378)
- trends does not plot global trends (see above)

Another thing @ghosh97 : please make sure that all ocean3d diagnostics write regions in output filenames (both figures and netcdf) consistently as "*.indian_ocean.pdf" and not "*.Indian_Ocean.pdf" (a `.lower()` needs to be added when passing the region name)

jhardenberg:
the 200hPa humidity bias is understandably empty since there is almost no humidity there ....
image

I understand the problem that we cannot suppress it or change the color range since we cannot distinguish them per level, unless we make separate config files...so I guess it is ok to keep it as it is.
@silviacaprioli what do you think?

silviacaprioli:
> I understand the problem that we cannot suppress it or change the color range since we cannot distinguish them per level, unless we make separate config files...so I guess it is ok to keep it as it is. @silviacaprioli what do you think?

As you said, the quickest fix would be to create two separate configuration files, `config-atmosphere3d-200hPa` and `config-atmosphere3d-850hPa`, and adjust the color ranges accordingly.
For a more long-term solution, we could consider improving the diagnostic itself to handle these cases,  maybe with `vmin` and `vmax` as lists of `len(levels)`?

jhardenberg:
What about simply leaving the limits for q in atmosphere3d biases automatic ? (I just inmplemented this)

jhardenberg:
Look here the atmosphere3d biases, I believe this is a good solution for this cycle, we leave q with automatic range. Then in the future it would be nice to implement what you suggest @silviacaprioli 
https://dev-climatedt-internal-dashboard.2.rahtiapp.fi/aqua

ghosh97:
> To summarize the situation:
> 
> * stratification is still not working (see comment in [Improvement of ocean stratification #2378](https://github.com/DestinE-Climate-DT/AQUA/pull/2378))
> * trends does not plot global trends (see above)

I've wrote a brief reply in https://github.com/DestinE-Climate-DT/AQUA/pull/2378#issuecomment-3449049145 
The issue is with the contour plot with global data when the lon range is 0 to 360.

tovaz92:
Hi, I had a quick look to https://github.com/DestinE-Climate-DT/AQUA/issues/2373 and since it was a quick thing I added it. I did also a quick refactor of the time function. I can remove it if create confusion.

jhardenberg:
@ghosh97 in trends I changed your 
```
for region, climatology in zip(regions, climatologies):
```
to
```
for region in regions:
                for climatology in climatologies:
```
because if regions=['go', 'po', 'ao', 'io']  and seasons=['January']  it would iterate only over 'go', 'January' . Was this your intention? I.e. should I have set seasons=['January', 'January', 'January', 'January'] ? This seems not very intuitive. With my solution it iterates over all combiantions of regions and seasons (but of cours eit does not allow finer selection)

jhardenberg:
@ghosh97 instead I had to remove these:
```
            # Add the global region if not present
            if regions != [None] or 'go' not in regions:
                 regions.append('go')
```
because it would lead to doing global twice (once because in the regions list and once because always added here. Better to be explicit.

jhardenberg:
They are all here: 
 /pfs/lustrep3/projappl/project_465000454/jvonhar/aqua-analysis/output/climatedt-phase1/IFS-NEMO/ssp370/r1
 
 I have rerun atmosphere2d-lat_lon_profiles and now it worked ....

oloapinivad:
> Let e rerun also drift which failed (I will save the current log as ocean3d-drift.log.1)

The error seems to suggest that something wrong is gong with region, probably the change of logic you did above breaks something....

jhardenberg:
> > Let e rerun also drift which failed (I will save the current log as ocean3d-drift.log.1)
> 
> The error seems to suggest that something wrong is gong with region, probably the change of logic you did above breaks something....

The frustrating thing is that of course yesterday I tested them all, so this was a surprise...

jhardenberg:
ok, I am rerunning it now (you can check the new log being generated there) 
The new log has an interesting error at the end

src/aqua/analysis/analysis.py
```diff
@@ -92,66 +108,77 @@ def run_diagnostic_func(diagnostic: str, parallel: bool = False, regrid: str = N
...
         loglevel (str): Log level for the diagnostic.
         logger: Logger instance for logging messages.
-        aqua_path (str): AQUA path.
         cluster: Dask cluster scheduler address.
     """
 
-    script_dir = config.get('job', {}).get("script_path_base")  # we were not using this key
-    if not script_dir:
-        script_dir=os.path.join(aqua_path, "diagnostics")
+    # safety check on clu
```
mnurisso: typo

src/aqua/util/time.py
```diff
@@ -236,18 +236,18 @@ def time_to_string(time=None, format='%Y-%m-%d'):
...
-        return pd.to_datetime(time).strftime(format)
+
+    # Convert supported types into pandas.Timestamp
+    if isinstance(time, (str, pd.Timestamp, np.datetime64)):
+        ts = pd.to_datetime(time, errors='coerce')
     else:
         raise ValueError('time_to_string() requires a time argument of type str, pd.Timestamp or np.datetime64')
 
+    if pd.isna(ts):
+        return 'N.A.'
```
mnurisso: I again advocate once against this change.
When time_to_string is called a description is generated. If I am for example calling it in the `std_startdate` and the value is None the issue is of the diagnostic not of the function.
The diagnostic has to be aware of the `std_startdate` and eventually fill the value or not call the function at all. Like this we will end up with description writing N.A. in the dashboard, I don't see how this is the result we want.

mnurisso (commented):
Left some comment

ghosh97:
> @ghosh97 in the config for stratification we currently have:
> 
> ```
>       regions: ['go', 'io', 'ao', 'so', 'arc_o', 'po']
>       climatology: ['January']
> ```
> 
> which will not work with zip because we need to list climatologies for each of them. Should I put January for all of them or is there another combination which would make sense from a scientific point of view. What should we focus on for ocean evaluation?

Hi @jhardenberg 
It should be like this:
```
      regions: ['ls', 'ros', 'is', 'ws', 'gs']
      climatology: ['DJF', 'DJF', 'JJA', 'JJA', 'DJF']
```

src/aqua/util/time.py
tovaz92: Ok lets revert this then, I'll just keep the condensed `isinstance` part.

jhardenberg:
Ok, I may have fixed all. 
I will now relaunch a full aqua analysis and disappear for 3h ( to give a lecture) ... let's see what happens. You can follow the logs in real time in the usual place

jhardenberg:
This is now broken 
image
@mnurisso

jhardenberg:
The pacific ocean trends are not plotted correctly:

image


I suggest to drop the pacific ocean from the target regions and to solve this separately in anothe rissue/PR

jhardenberg:
Also the Ross sea MLD is all worng, so I will drop it for now

image


@ghosh97

jhardenberg:
@silviacaprioli I fixed the limits discussed above

jhardenberg:
I think that we should not break the format of the changelog and anyway make on entry line (even if long):
I add below each line acomment or more compact version and at th end I would join them all together:

> * Diagnostics reorganised into meta-diagnostics categories (atmosphere2d/3d, ocean etc.) calling diagnostics tools, implemented in new config files in `/config/analysis`

--> "Refactoring of aqua analysis to allow config file with diagnostic groups and reorganization of diagnostic config files. Optimizations of diagnostic parallel execution."

> * AQUA analysis refactoring with command line parser and execution logic

??? We did no add the CLI here, we just refactored the  config and the ability of aqua analysis to process it, so the line above should be enough

> * Added global task cluster support for parallel diagnostic execution

??? this we had before, we did not add it here

> * Improved logging with per-diagnostic log files and error reporting

??? We already had that before

> * Ocean output and plotting refactoring: Enhanced multi-format saving management with descriptive metadata and implementation with AQUA metadiagnostics support

This was done in other PRs merged here. We just fixed some descriptions and small bugs here.
Maybe we can add "including minor fixes to output filenames and  figure descriptions and color ranges"

To summarize, I would add a single line like this:

```
Refactoring of aqua analysis to allow config file with diagnostic groups and reorganization of diagnostic config files. Optimizations of diagnostic parallel execution. Inlcuding minor fixes to output filenames, figure descriptions and color ranges. 
```

mcadau:
Ok I like it! Let's keep it on a single line. Yours is indeed comprehensive. 
I set the commit for now, if others want to change we fix it. In this way if you want to merge, we're free to go.

And about the test that were failing on LatLonProfiles, maybe it was because of the filename? Now that you're running tests LatLonProfiles and PlotLatLonProfiles went well.

jhardenberg:
We still have this unresolved conversation
image

jhardenberg:
Currently `push_analysis` is independent on any external env variable definition, and we can always pass `--config` (this is also what the worflow does)

jhardenberg:
Ok, let's merge, the discussion above, if important, we can solve separately.

Issue #2377, opened 2025-10-24
Ocean3d stratification failing because coordinate must be equally spaced


jhardenberg:
### machine

Levante

### catalog

ci

### version

main

### What happened?

Ocean stratification is failing with the error below. 
The config file is `/users/jvonhar/.aqua/diagnostics/ocean3d/config-ocean3d-en4.yaml`

Due to the code freeze this is quite urgent, thank you.

### Code to reproduce the issue

```python
python /pfs/lustrep3/projappl/project_465000454/jvonhar/AQUA/src/aqua_diagnostics/ocean_stratification/cli_ocean_stratification.py --model IFS-NEMO --exp ssp370 --source lra-r100-monthly --outputdir /pfs/lustrep3/projappl/project_465000454/jvonhar/aqua-analysis/output/climatedt-phase1/IFS-NEMO/ssp370/r1/ocean3d --catalog climatedt-phase1 --realization r1 --config /users/jvonhar/.aqua/diagnostics/ocean3d/config-ocean3d-en4.yaml -l DEBUG
```

### Error messages

2025-10-24 19:41:14 :: OutputSaver :: DEBUG    -> Available metadata: {'diagnostic': 'ocean3d', 'diagnostic_product': 'stratification', 'catalog': 'obs', 'model': 'EN4', 'exp': 'v4.2.2', 'realization': 'r1', 'region': 'Atlantic_Ocean', 'timestamp': '2025-10-24 19:41:14', 'aqua_version': '0.18.1'}
2025-10-24 19:41:22 :: OutputSaver :: INFO     -> Saved NetCDF: /pfs/lustrep3/projappl/project_465000454/jvonhar/aqua-analysis/output/climatedt-phase1/IFS-NEMO/ssp370/r1/ocean3d/netcdf/ocean3d.stratification.obs.EN4.v4.2.2.r1.Atlantic_Ocean.nc
2025-10-24 19:41:22 :: Stratification :: INFO     -> NetCDF file saved successfully.
2025-10-24 19:41:22 :: Stratification :: INFO     -> Stratification diagnostic saved to netCDF file.
2025-10-24 19:41:22 :: PlotStratification :: DEBUG    -> Variables in data: ['mld']
[...]

PR #2378, opened 2025-10-25
Improvement of ocean stratification and trends

ghossh:
Close https://github.com/DestinE-Climate-DT/AQUA/issues/2377

- Updating the CLI of stratification to read multiple climatologies. 
- Improvement of tests in ocean stratification
- compute data in ocean trend for faster plot
- convert longitude range from 0 to 360 degree to -180 to 180 degree in trend and MLD diagnostics.

### DIFF
--- a/src/aqua_diagnostics/ocean_trends/plot_trends.py
+++ b/src/aqua_diagnostics/ocean_trends/plot_trends.py
@@ -162,8 +162,6 @@ class PlotTrends:
 
     def set_data_list(self):
         """Prepare the list of data arrays to plot."""
-        self.logger.debug('Loading the data into memory')
-        self.data = self.data.compute()
         self.data_list = []
         if hasattr(self, "levels") and self.levels:
             self.data = self.data.interp(level=self.levels)


src/aqua_diagnostics/ocean_trends/plot_trends.py
```diff
@@ -149,6 +162,8 @@ def set_ytext(self):
 
     def set_data_list(self):
         """Prepare the list of data arrays to plot."""
+        self.logger.debug('Loading the data into memory')
+        self.data = self.data.compute()
```
jhardenberg: Please remove these two lines, we deal with loading the data in #2371. Let's wait to have all together and then we can make further adjustments.

mnurisso:
I am sorry to go back to this PR but the values in test are totally different from the previous tests done by the previous diagnostics, this sound quite strange to me, expecially the rho value which went from 26 to -3*10^8.
To me, even if the figures produced by the cli are fine, I strongly advise to check tests after the freeze

ghosh97:
Hi @mnurisso 

I've already checked the values; the values are the same as before.
This is from the old stratification test: https://github.com/DestinE-Climate-DT/AQUA/blob/aba786075fa6b37426cbbde27f4c8798668c4a63/tests/ocean3d/test_ocean_circulation.py#L77

this one, the refractored one:
https://github.com/DestinE-Climate-DT/AQUA/blob/aba786075fa6b37426cbbde27f4c8798668c4a63/tests/ocean3d/test_stratification.py#L23

the values are different because of its from differeent region/points, but the scale is same.

PR #2380, opened 2025-10-27
chunking at Reader() level for non netcdf sources

oloapinivad:
## PR description:

With @jhardenberg we noticed that chunking is not passed when the source is a netcdf at reader level to intake. This is likely because the `filter_kwargs` does not allow it to reach. A temporary solution, before revisiting entirely the code toward the intake2 transition is to add it at the list of argumets to be not filtered. 

 - [x] Changelog


oloapinivad:
The 6 tests failing are for catgen and stratification, I guess this one is ok on the other hand

jhardenberg:
Looks ok to me and the failing tests are the catgen ones...so merge ? (this way I can continue in the other PR)

PR #2386, opened 2025-10-28
Documentation update

mnurisso:
## PR description:

PR to update the documentation that may have changed in the latest PR for the freeze


jhardenberg:
Now that in #2371 we have introduced the distinction between diagnostics and tools, maybe should we mention this also in the docs?

PR #2387, opened 2025-10-28
Ensemble for single model timseries feedback

mnurisso:
## PR description:

Modifications needed to run the timeseries ensemble analysis for a single model automatically with the CLI.

I have added a modification that reads from intake the available realizations so that it is not needed anymore as user input.

Close #issue_number

## People involved:

----

 - [ ] Changelog is updated.

### DIFF
--- a/config/diagnostics/ensemble/config_single_model_timeseries_ensemble.yaml
+++ b/config/diagnostics/ensemble/config_single_model_timeseries_ensemble.yaml
@@ -7,16 +7,7 @@ datasets:
   - catalog: 'climatedt-o25.1'
     model: 'IFS-NEMO'
     exp: 'hist-1990-ensemble'
-    source: 'aqua-ocean-timeseries-levante'
-    realization: ['r2', 'r3', 'r4']
-    regrid: null
-
-# Reference
-references:
-  - catalog: null  
-    model: null
-    exp: null
-    source: null
+    source: 'aqua-ocean-timeseries'
     regrid: null
 
 


src/aqua_diagnostics/ensemble/cli_single_model_timeseries_ensemble.py
```diff
@@ -99,7 +123,12 @@ def parse_arguments(args):
                         exp = get_arg(args, "exp", dataset[0]["exp"])
                         source = get_arg(args, "source", dataset[0]["source"])
                         regrid = get_arg(args, "regrid", dataset[0]["regrid"])
-                        realization = get_arg(args, "realization", dataset[0]["realization"])
+                        realization = extract_realizations(
```
oloapinivad: this is using all available realization by default, correct? wouldn't make sense to leave as an option to specify WHICH realization the user want to process?

mnurisso: the realization parameter by the cli arg does not make sense because it can be only one. The dataset[0] can be an option as you mention and I can change it like this.

mnurisso: Yes but now we can handle this case with the catalog. Furthermore the CLI do not fail if a file is not found for a realization but that realization is skip

maqsoodrajput (approved):

PR #2361, opened 2025-10-21
Use fixer or retrieve data to reduce test time

tovaz92:
This PR aims to optimise, where possible, the execution times of certain tests by using a set of common `Fixture` and avoiding duplicated data retrieve throughout the tests.

----

 - [x] Changelog is updated.

Issue #1481, opened 2024-11-19
Race condition in aqua-analysis in weight generation


jhardenberg:
### machine

Levante

### catalog

nextgems4

### version

None

### What happened?

In doing tests with aqua-analysis on levante (since lumi is offline these days) I got a new error for which some diagnostics may not run.
What happens is that both tropical precipitation and global means diagnostics for some reason decide to regrid the LRA data to the same grid (r100). This may be debatable and removed in future versions, but the problem is another one: 
when they start they both realize that on this machine the specific weight files for this transformation are missing and ... they try at the same time to reconstruct them (the diagnostics run in parallel threads and we have multiple instances of the diagnostic for different models running  at the same time). This is doomed to fail since there are two weight generation processes trying to write to the same file at the same time.

I am not sure how this could be solved. for me it highlights the risk also in the future that by chance (I admit you have to be unlucky) two instances of the workflow may try to access in write the same files at the same time  

### Error messages

2024-11-18 23:57:12 :: Reader :: WARNING  -> Weights file not found: /work/bb1153/b382075/aqua/weights/weights_IFS-FESOM_historical-1990_lra-r100-monthly_ycon_r100_l2d.nc
2024-11-18 23:57:12 :: Reader :: WARNING  -> Attempting to generate it ...
2024-11-18 23:57:12 :: Reader :: DEBUG    -> Grid size comparison - Original: 64800, New: 64800.
2024-11-18 23:57:12 :: Reader :: DEBUG    -> The total expected processing time is 0.5729823664707385 seconds.
Traceback (most recent call last):
  File "/home/b/b382075/mambaforge/envs/aqua/lib/python3.12/site-packages/xarray/backends/file_manager.py", line 211, in _acquire_with_cache_info
    file = self._cache[self._key]
           ~~~~~~~~~~~^^^^^^^^^^^
  File "/home/b/b382075/mambaforge/envs/aqua/lib/python3.12/site-packages/xarray/backends/lru_cache.py", line 56, in __getitem__
    value = self._cache[key]
            ~~~~~~~~~~~^^^^^
[...]

PR #2402, opened 2025-10-30
Parallelise with pytest-xdist and add centralised fixtures

### LINKED ISSUE #2395: Tests: parallelisation with xdist to further reduce time
### Discussion

This issue aim to discuss other possible improvements to reduce the execution time for tests:

#### Parallel tests with [pytest-xdist](https://pytest-xdist.readthedocs.io/en/stable/#pytest-xdist)

There might be a possible way to parallelise the execution of tests using this package, based on `pytest`.
What do you think of this?

---

_Update 31/Oct/25:_

As the parallelisation is working (https://github.com/DestinE-Climate-DT/AQUA/pull/2402/commits/2666f59ccdbf68a860e80fe53745444b8a45a481), a clean way to choose what test should run in parallel and what sequentially, is to use  `pytestmark` with `xdist_group` defined at the module level:
- Pytest automatically applies these markers to every test in that module (functions, classes, and all methods in classes).
- It’s equivalent to adding `@pytest.mark.aqua` and `@pytest.mark.xdist_group(name="dask_operations")` to every test.
- Markers accumulate: if a test already has a decorator, they both apply (no conflicts).

When pytest-xdist sees tests with `@pytest.mark.xdist_group(name="group_name")` marker,  
(1) all the tests with the same `name="<group_name>"` will run in the same worker, avoiding any dask cluster conflict.  
(2) tests with different groups will run in parallel on different workers, and  
(3) tests without groups will be distributed to the free workers.

tovaz92:
## PR description:

1. This PR start from changes in #2361 which aims to optimise, where possible, the execution times of tests by using a set of common Fixture in a new `conftest.py` file which avoid duplicated data `retrieve` throughout the tests. 
Further changes have been applied to centralise the loglevel and dpi.

2. Test for implementation in GitHub workflow of parallel testing with 'pytest-xdist',  reducing execution time of tests by ~50%.  

Close #2395 

----

 - [x] Changelog is updated.
 - [x] pyproject.toml is updated if needed


tovaz92:
The tests succeeded! And we achieved a great execution time reduction of about 50%

tovaz92:
The `tests/test_ocean3d/test_stratification.py` is failing with the following error:

```python
src/netCDF4/_netCDF4.pyx:2521: in netCDF4._netCDF4.Dataset.__init__
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
>   ???
E   PermissionError: [Errno 13] Permission denied: '/__w/AQUA/AQUA/AQUA/AQUA_tests/weights/weights_test-hpz3-nested-fesom-3d_ycon_r100_l3d-level.nc'
src/netCDF4/_netCDF4.pyx:2158: PermissionError
```

As this test did not fail before in the run I was doing earlier, and it did not changed, copilot was suggesting that there might be a problem related to multiple pytest-xdist workers trying to generate/write/accessing the same weights file simultaneously.

@ghosh97 do you have any advice/opinion on what might cause this?  
A possibility could be to run stratification sequentially and not in parallel, but it would be good to know if you are aware on any Permission Problem that could be a quick solution.

jhardenberg:
Well, maybe the test writing the file should also be the one testing it later ... so avoid dependencies between tests

tovaz92:
Test are passing! In latest https://github.com/DestinE-Climate-DT/AQUA/pull/2402/commits/aa5a50eef372c929ac29055563cb253c87846e9c commit,  `filelock` is introduced to avoid race conditions in`_load_area()` and `weights()` (when netcdf file are written simultaneously). 
This solution should solve possible overlaps and race condition in the long-term across all the tests (and future ones) for files.  
If somebody could have an extra check in reviewing these last changes and verify they are indeed ok, this would be appreciated.
I will re-trigger test now for final check.

jhardenberg:
It is indeed very good to have complex tests running in parallel, because this way we test everything in a more realistic parallel envirionment.

I do not like much that the change on the race condition is done here, it should be in a separate PR, also because we have an open issue on that https://github.com/DestinE-Climate-DT/AQUA/issues/1481
Also it would be cleaner if this PR modified basically only the actions and tests.

I would be very careful with filelocks. We use them for the yaml files with the catalog generator because in that case we are modifying an existing file. Clearly two process cannot modify a file at the same time because modifying means reading the file, applying a change and rewriting it and if two processes do that that could be a disaster.

The situation with area and weight generation is very different because we always recreate the same file, so indeed this could be done at the same time, but also in this case two processes should not do that at the same time because they will partially overwrite the other's file. The solution is simpler than using a lock: write the file first to a temporary filename, then do a mv to the final file (which is instantaneous and the filesystem avoids any problems)

Should we try to implement maybe first this solution i a separate PR ? I can probably do it quickly, so you could test it also here (merging into a subbranch)

jhardenberg:
@tovaz92 I tested what I describe above in #2418, which is a subbranch of this one + merging the branch of #2412.
The good news: you parallel tests all worked nicely and in just 13 minutes. Congratulations, a great result! 

So to proceed in an orderly way: if we merge #2412 into main then you can merge main here (after reverting  aa5a50eef372c929ac29055563cb253c87846e9c !) and this PR will be ready

environment.yml
```diff
@@ -22,6 +22,7 @@ dependencies:
   - xesmf # for ECmean
   - intake-esm
   - imagemagick
+  - pytest-xdist # for parallel test execution
```
oloapinivad: you should add it only to the pyproject, no need to add it here, this is for binaries

tovaz92: ok thanks!

tests/seaice/test_plot_seaice.py
```diff
@@ -48,14 +55,14 @@ def setup_class(cls):
 
     def test_check_as_datasets_list_valid_input(self):
         """Validate the _check_as_datasets_list utility with valid input."""
-        plot_seaice = PlotSeaIce()
+        plot_seaice = PlotSeaIce(dpi=DPI)
```
jhardenberg: Why is there all this DPI stuff here, is this not something from #2361  ?

jhardenberg:
I am a bit worried about the overlap with #2361 which also was touching DPIs , will this not lead to conflicts? @tovaz92

oloapinivad:
> I am a bit worried about the overlap with #2361 which also was touching DPIs , will this not lead to conflicts? @tovaz92

isn't this PR superseded by the current one?

src/aqua_diagnostics/global_biases/plot_global_biases.py
```diff
@@ -78,7 +78,7 @@ def _save_figure(self, fig, diagnostic_product,
                                  extra_keys=extra_keys, metadata=metadata)
         elif format == 'png':
             outputsaver.save_png(fig, diagnostic_product=diagnostic_product,
-                                 extra_keys=extra_keys, metadata=metadata)
+                                 extra_keys=extra_keys, metadata=metadata, dpi=self.dpi)
```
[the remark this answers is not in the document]
jhardenberg: Yes, but
a) it is not in the PR description
b) isn't the same thing done in #2361 ?

tovaz92:
Yes, all the old changes from https://github.com/DestinE-Climate-DT/AQUA/pull/2361 are in this PR as well. I was keeping on-hold https://github.com/DestinE-Climate-DT/AQUA/pull/2361 and only working on this PR. I'll close the old one and update the changelog

tovaz92:
As tests are passing after these extra refinement, now I will remove the two-actions temporary solution by substituting the parallelised tests into the original `aqua_test` job.

tests/seaice/test_plot_seaice.py
tovaz92: Many of these changes come from #2361, but in this PR I centralised the value setting a `DPI=50` in conftest, which can be imported directly. I also incremented the number of tests that can use DPI in this PR.

mnurisso (approved):

jhardenberg:
I would merge this so we can use it to test #2445

tovaz92:
I would wait a bit as we are experiencing fails when the tests are execute locally, although they work fine when run on the GItHub jobs workflow.

jhardenberg:
These are mine
image

The rounding errors can all be fixed with some more tollerance.  
Those with /app could  be avoided if the catalog for ci had `FDB_HOME: ./AQUA_tests/fdb` instead of `FDB_HOME: /app`

mnurisso:
I think, as said to @tovaz92 that the rounding error is due to some small changes done in the DVC and missing in the dkrz backup.
So I don't think the issue is to increase the tolerance, but to find a way to have a dvc download that users can do, otherwise it will not be possible to replicate tests locally (and regrid whatever dataset)

PR #2408, opened 2025-10-31
Fix exposed token in tests

oloapinivad:
## PR description:

A recent PR introduced a probably unnecessary feature that exposes the token of aqua actions, which implies that the token gets invalidated.  This is testing the reverting.

## People involved:
@tovaz92


tovaz92 (approved):
approving

PR #2410, opened 2025-10-31
Add r100 grid for lonlat with a grid file

### LINKED ISSUE #2405: Provide a grid file for r100 lon-lat
Currently our log files are full of lines like these because the lon-lat grid used by LRA data has not grid file and a `_retrieve_plain()` is needed (this one is from atmosphere2d-timeseries in a aqua analysis run):

```
2025-10-30 09:22:24 :: Regridder :: WARNING  -> Source grid path not found. Please provide a dataset.
2025-10-30 09:22:24 :: Reader :: WARNING  -> Issues in the Regridder() init: trying with data
2025-10-30 09:22:24 :: Reader :: DEBUG    -> Getting sample data through _retrieve_plain()...
2025-10-30 09:23:05 | smmregrid.GridInspect |     INFO -> Updating the log_level to DEBUG

```

So about 40s are wasted loading info on a grid we know all about. This time is multiplied by all figures produced, increasing consistently the execution time. 
I suggest to provide a grid file in dvc (we can just copy the existing `cell_area_r100.nc` file) and add an entry in `grids/default.yaml`like this:

````
  lon-lat-r100:
    path:
       2d: '{{ grids }}/grids/lonlat/r100_grid.nc'
    space_coord: [lon, lat]
````

and finally update the catalog  (and the LRA catalog generator) to use `source_grid_name: lon-lat-r100` for LRA data

jhardenberg:
## PR description:

This adds an explicit grid file for the r100 lon-lat grid (the one we use for LRA). 
This leads spare a lot of wasted time when running diagnostics because we do not need to do a `_retrieve_plain()`

The file `grids/r100_grid.nc` has already been added to dvc.  

We need to fix the LRA catgen in this PR so that it uses `source_grid_name: lon-lat-r100`
and then update the catalogs to use this (but this way we would make the catalogs incompatibel with older AQUA versions ... ?)

Close #2405 

----

 - [x] Changelog is updated.
 - [x] catgen updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,9 +15,13 @@ ClimateDT workflow modifications:
 
 AQUA core complete list:
 - Use grid with explicit grid file for lat-lon-r100 (#2410)
+- Realization formatting correctly processed by the Reader (#2392) 
+- Realization (and other intake kwargs) are an attribute of the Reader (#2392)
+- Refactoring of aqua analysis to allow config file with diagnostic groups and reorganization of diagnostic config files (#2371)
+- Optimizations of diagnostic parallel execution (#2371) 
+- Including minor fixes to output filenames, figure descriptions and color ranges (#2371)
 - Enumerate aqua analysis log file for multiple config files (#2407)
 - Fix push_analysis.sh rsync functionality to not use ssh (#2403)
-- Refactoring of aqua analysis to allow config file with diagnostic groups and reorganization of diagnostic config files. Optimizations of diagnostic parallel execution. Cleaner distinction between `diagnostics` and `tools` (#2371)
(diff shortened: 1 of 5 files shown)


oloapinivad:
One detail you might remember, it is very important to modify the LRA catalog generator otherwise this will never be used.  No need to modify older catalogs.

jhardenberg:
As agreed with @oloapinivad  changing only climatedt-phase1/IFS-NEMO/ssp370 in the catalogue to allow this to be included in aqua-analysis stack testing in the future https://github.com/DestinE-Climate-DT/Climate-DT-catalog/pull/207

jhardenberg:
uodated changelog

PR #2412, opened 2025-10-31
Parallel safe creation of weights and areas

jhardenberg:
## PR description:

As discussed in #1481 we are not creating weight and areas files safely because parallel diagnostic processes (e.g. launched by aqua anlysis) could happen to try and do that at the same time.
This has also been discussed in #2402 where it leads to failure of parallel tests.

This implementsa safe approach where temporary files are created first on the same filesystem and then renamed (with a replace operation) . This is istantaneous and the filesystem/OS takes care of any conflict.

Close #1481 

 - [x] Changelog is updated.

PR #2418, opened 2025-11-02
[Test] Test of Parallel pytest xdist

jhardenberg:
Just a temporary test of #2402  (PArallel pytest with xdist) including #2412 (safe weight and area creation). Not to be merged, just to see if tests run


jhardenberg:
I did not realize that closing this test would also close the test results but all worked well and the results are available here: 
https://github.com/DestinE-Climate-DT/AQUA/actions/runs/19004730261

PR #2422, opened 2025-11-04
Remove useless lower and replace for region in diagnostics

jhardenberg:
## PR description:

This removes a lot of useless replace() and lower() to force region to desired format which is already done by OutputSaver

----

 - [x] Changelog is updated.


jhardenberg:
Tested by running also a full aqua-analysis stack

PR #2425, opened 2025-11-05
Centralise core figure save function

### LINKED ISSUE #2109: Homogeneisation of save_fig functions
### Feature request

Hi all, after a discussion with @silviacaprioli we realised the existence of at least three cases in which a `save_plot` function was defined internally of the relative diagnostics: 

- [Case 1 - timeseries](https://github.com/DestinE-Climate-DT/AQUA/blob/main/src/aqua_diagnostics/timeseries/base.py#L392)
- [Case 2 - global bias](https://github.com/DestinE-Climate-DT/AQUA/blob/d03755995144f94514e63d2ac9148acec6b50a7a/src/aqua_diagnostics/global_biases/plot_global_biases.py#L35)
- [Case 3 - sea ice](https://github.com/DestinE-Climate-DT/AQUA/blob/567f40162b105c3efc394934e7fe15064563433a/src/aqua_diagnostics/seaice/plot_2d_seaice.py#L477)

As you can see, each of these functions have very minor differences among each other.  
This is a repetition that might be needed but maybe it is avoidable. 

I raise here a discussion in case we want to tackle this as a group, and what your impressions are about this.

tovaz92:
Centralise the save function to avoid repetition of wrappers.

Close #2109

----

 - [x] Changelog is updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ ClimateDT workflow modifications:
 - New position for push_analysis ``config.grouping.yaml` in `$AQUA/config/analysis`
 
 AQUA core complete list:
-- Centrlise `save_figure` in OutputSaver class (#2425)
+- Centralise `save_figure` in OutputSaver class (#2425)
 - Allow longer time default for connection to dask cluster (#2420)
 - Safe parallel creation of area and weight files (#2412)
 - New grid lat-lon-r100 with explicit grid file, new default in DROP catgen (#2410, #2413)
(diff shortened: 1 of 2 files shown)


silviacaprioli:
That's a good idea, since a lot of diagnostics have a customized save_figure method! Are you planning to update those diagnostics to use it in this PR?

tovaz92:
@silviacaprioli yes, I tried to do this in the last commit, let's see how the tests goes.

src/aqua_diagnostics/core/output_saver.py
```diff
@@ -445,3 +445,76 @@ def _create_catalog_entry(self, filepath, metadata, jinjalist=None, wildcardlist
 
         self.logger.debug("Releasing catalog file %s", catalogfile)
         return catblock # using this in the tests
+
+
+def save_figure_core(fig: Figure,
```
silviacaprioli: isn't it better to put it in core/diagnostic.py instead of output_saver.py?

PR #2431, opened 2025-11-05
[LatLonProfiles] Lines plot sorting

### LINKED ISSUE #2427: LonLat profiles plot first reference then model
### machine

Lumi

### catalog

climatedt-phase1

### version

main

### What happened?

In plots like this

Image

it would be better to plot the blue line (the model) over the black one and not viceversa, so it should be the second one to plot

```python

```

### Error messages

_No response_

None

mcadau:
## PR description:
As discussed in #2427, the plotting order for model and reference line is reversed, so that model line is not overwritten by the reference line.

Close #2427 

----

 - [x] Changelog is updated.


jhardenberg:
> @jhardenberg I here address the issue you found in #2427. Do you think that the multiple line profiles (check [this](https://github.com/DestinE-Climate-DT/AQUA/blob/main/notebooks/diagnostics/lat_lon_profiles/multiple_lines_profiles.ipynb) notebook) should be updated accordingly? If i'm not wrong they're not currently part of aqua analysis, but maybe they could be useful in the future

The multiple lines are cool and  indeed they will be useful in the future. Yes, also for them I would plot the lines over the reference

mnurisso:
FYI: there is a zorder arg in most of plot routine to order them without having to change where the code to plot them is placed inside the functions

mnurisso (approved):

PR #2432, opened 2025-11-05
Filelock in drop

### LINKED ISSUE #2429: LRA_GENERATOR jobs fail intermittently
### machine

MN5

### catalog

ci

### version

v0.18.1

### What happened?

See full GitLab Issue: [1340](https://gitlab.earth.bsc.es/digital-twins/de_340-2/workflow/-/issues/1340)

@jhardenberg did you encountered this? 

if the issue is a run condition, a possible solution could be to add a `FileLock` in `create_catalog_entry` in `catgen.py`.

Maybe

jhardenberg:
## PR description:

This should solve #2429 by adding a filelock when changing the catalog source yaml files in drop. We really need to make sure that anytime we read , change, and rewrite a file this is done in a thread-safe / parallel safe way. I will check if there are other places. Notice: Instead just generating a file (but not reading it first because we do not modify it) does not require a lock but rather writing to a temp file and then renaming, as we do for areas and weights now.

close #2429 

----

 - [x] Changelog is updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ ClimateDT workflow modifications:
 
 AQUA core complete list:
 - Add filelock when modifying catalog sources (#2432)
+- Area selection full support for regions across the dateline (#2430, #2434)
 
 AQUA diagnostics complete list:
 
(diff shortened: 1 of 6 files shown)


jhardenberg:
There were some load_yaml/dump_yaml sequences also in the aqua console, now with lock. It is extremely unlickely that two job scripts may do this at the same time (it is actually something users are supposed to do) ... but better safe than sorry

oloapinivad:
Cool, this looks good! Only one question: how would this filelocking work with intake? This will be applied by our internal writing operations from drop, but what if a filelock is active while intake is trying to read the source? I am just worried that this will not have effect there

Issue #2399, opened 2025-10-30
Ocean plots with issues for some regions + season in MLD plot titles


jhardenberg:
### machine

Lumi

### catalog

climatedt-phase1

### version

main

### What happened?

As first reported in #2371, Ocean trends does not plot correctly the PAcific Ocean, presumably because it crosses the dateline:

<img width="799" height="281" alt="Image" src="https://github.com/user-attachments/assets/bfa03ef3-6f7a-4cad-bfd4-3089e7843a0e" />

Same thing for MLD in Ocean stratification. The Ross Sea (which is across the date line) is plotted like this:

<img width="816" height="68" alt="Image" src="https://github.com/user-attachments/assets/6477fc43-a683-4e6f-b0f1-0141743c52cb" />

### Code to reproduce the issue

```python

```

### Are you interested in making a pull request?

None

jhardenberg:
Another thing: since the season of the analysis  is an important information, it should be included in the plot title and in the description !

jhardenberg:
@ghosh97 we are waiting to merge #2371 so please wait for that and let's work based on the main

jhardenberg:
Ok, merged, so now we can work based on main. The plot title and description seems to me actually something to be fixed urgently

jhardenberg:
@mnurisso could you give an opinion?

mnurisso:
2) the issue is not in the plot_single_map function provided by the framework, see code below and plot

<img width="732" height="641" alt="Image" src="https://github.com/user-attachments/assets/7a85f954-f78b-48ff-99d1-48cccbc4ea02" />

3) therefore the issue is in the custom function used by the diagnostic, the `plot_maps` available which is not part of the framework

PR #2433, opened 2025-11-06
Fixing lat lon plot in ocean3d (ocean trends and MLD)

ghossh:
This PR is addressing the issue of Ocean plots with issues for some regions. 
After introducing https://github.com/DestinE-Climate-DT/AQUA/pull/2430, I've stopped conversions of longitude and added central longitudes in the projection.

Close https://github.com/DestinE-Climate-DT/AQUA/issues/2399


mnurisso:
Can you please modify the config files accordingly adding back the regions that were removed due to the plot issue? Ross Sea for MLD and Pacific for multi level trend

jhardenberg:
This worked well with aqua analysis, we can merge!

jhardenberg:
The only thing that confuses me a bit: in `config/tools/ocean3d/definitions/regions.yaml` the Ross Sea longitudes are specified as `[160, -150]`, while other regions have for example `[300, 316]` (Labrador Sea). Will both a 0:360 and a -180:180 syntax work?

jhardenberg:
Another observation: The MLD plots do not have anything on their axes (no lon and lat ticks)

mnurisso:
> Maybe the description could remind us of the lon and lat limits of the domain then?

@ghossh while I'll merge this to backport it to v0.19, could you tackle this extra issue? I'll open an issue about this.

PR #2436, opened 2025-11-06
Remove warnings during tests

tovaz92:
Remove warnings such as:  `DeprecationWarning`, `FutureWarning`, `UserWarning:` ...  which appears during the tests execution.

Some warnings are removed by adapting `smmregrid` code in PR https://github.com/jhardenberg/smmregrid/pull/54

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ Unreleased in the current development version (target v0.20.0):
 
 AQUA core complete list:
 - Remove some warnings and adapt the code to the incoming standards (#2436)
+- Fix DROP CLI support for enddate, startdate and catalog (#2472)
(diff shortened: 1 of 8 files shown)


jhardenberg:
I am not sure that I understand entirely the goal of this PR @tovaz92 :
are you 'hiding' the warnings or changing the code of the tests so that they do not appear anymore? 
If it the second one, should we not do the same with our main code? A FutureWarning is a warning that something may break at the next package version update .... we should get rid of it addressing the ultimate cause

oloapinivad:
I think @tovaz92 is actually upgrading our code to the new standards of the different packages, whenever possible, so that `FutureWarning` are no longer there. 
It is impossible we do NOT suppress warnings in tests since these are fundamental to catch code changes

PR #2437, opened 2025-11-06
[Op] Filelock in drop - draft

jhardenberg:
## PR description:

This is the operational version of #2432 adding a filelock to prevent ruining yaml files from the catalog while modifying them

 - [x] Changelog is updated.


jhardenberg:
@mnurisso iof you do not like this we can do something with a cleaner approach ....

PR #2443, opened 2025-11-08
Safe dumping to yaml

jhardenberg:
## PR description:

Motivated by  #2429 and other current workflow issues, this (while not solving that problem entirely), addresses one remote possibility: if intake tries to read an experiment source while this is being dumped by another process a partially processed file might be read. This applies the same technique used also for areas and weights (first writing to a temporary file) to create the destination file atomically (and safely) with a `os.replace`...
Notice that the technique used (involving a temp directory instead of a temp file) should be safer and more robust.

If this works and passes tests we will have to tronsform this also into a fix for the operational branch.

----

 - [x] Changelog is updated.

### DIFF
--- a/src/aqua/util/yaml.py
+++ b/src/aqua/util/yaml.py
@@ -1,5 +1,6 @@
 """YAML utility functions"""
 
+from fileinput import filename
 import os
 from string import Template as DefaultTemplate
 from jinja2 import Template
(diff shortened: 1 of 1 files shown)


jhardenberg:
I am currently running this with a parallel aqua analysis on 3 ensemble members... let's see if it solves at least the "NO YAMLD data" issue

PR #2360, opened 2025-10-21
Introduce a `DiagnosticCLI` centralized class to avoid redundancy 

oloapinivad:
## PR description:

This is a nice tentative with a couple of prompts to Claude to introduce a centralized DiagnosticCLI to share the base features of each AQUA cli. The implementation looks nice, it can be simplified. Of course, it has to be tested and debugged yet. 

 - [ ] Tests are included if a new feature is included.
 - [ ] Documentation is included if a new feature is included.
 - [ ] Docstrings are updated if needed.
 - [ ] Changelog is updated.
 - [ ] Notebooks which requires changes are updated. 
 - [ ] environment.yml and pyproject.toml are updated if needed, together with the lumi installation tool. 
 - [ ] Diagnostic config files path are updated in the console if needed

close #2312

PR #2470, opened 2025-11-14
[Seaice] CLI refactor

tovaz92:
This PR refactor Sea ice CLI accounting for the latest changes and harmonise to the centralised CLI class.

Link to https://github.com/DestinE-Climate-DT/AQUA/pull/2360#issuecomment-3527067707

----

 - [x] Changelog is updated.


oloapinivad:
I still feel we are a bit out of target, sorry for being picky. My idea is to go into the direction of homogenizing as much as we can: the other diagnostics are creating a configuration file for each dataset that is used. So I would design it to have a two configuration files, one for the volume and one for the sea ice concentration. 
There is no need imho to have one config for each tool, also considering everything is meant to be work together as "sea ice". 
Also `startdate` and `enddate` should be part of the datasets, and not extra features.

Issue #1367, opened 2024-09-26
micromamba < 2.0 to unpin in actions


mnurisso:
### Discussion

We've pinned micromamba < 2.0 in #1365 due to https://github.com/mamba-org/setup-micromamba/issues/225

That's a backlog to not forget to unpin it when solved.
We've also set the cache to false, to be check if it can be reverted as well.

PR #2472, opened 2025-11-16
Fix DROP CLI support for enddate, startdate and catalog

### LINKED ISSUE #2471: DROP ignores --enddate
### machine

Lumi

### catalog

ci

### version

main

### What happened?

Investigating the latest workflow tests on MN5 (/home/pdt/pdt081320/wf/t0bn/LOG_t0bn) I realized that drop apparently is not considering `--enddate` so that we still have overlapping LRA instances.

So I tried the test below (which should only do the first month) and indeed it processes all 7 available months, instead of only the first one!

```python
cat > test.lra.cfg << EOF
target:
  resolution: r100
  frequency: monthly
  catalog: "ci"
paths:
  outdir: "/users/jvonhar/work/lra/LRA/"
  tmpdir: "/users/jvonhar/work/lra/tmp"
  configdir: "/users/jvonhar/.aqua"
options:
  loglevel: DEBUG
# all the variable listed here will be processed by the LRA creator
data:
  "IFS":
    "test-tco79":
      "long":
        workers: "1"
        vars: ['2t']
        realizations: [1]
EOF

aqua drop -c ./test.lra.cfg -d --realization r1 --enddate 20200131
```

### Error messages

_No response_

None

jhardenberg:
## PR description:

This addresses #2471 and fixes the CLI of DROP to honor enddate, startdate and catalog plus the following small fixes:

- Includes an additional cleanup of how the cli treats command line arguments over config file ones 
- adds a couple of missing zarr-related cli options. 
- Treats a rare case when enddate and startdate would create an empty dataset.
- Adds support to the reader for enddate and startdate also for non-fdb sources (required for testing)
- be more relaxed in dates accepted (also YYYYmmdd is now valid - the workflow uses this format)

If we approve this we will need to also make an OP version, since this functionality is required by the workflow (see gitlab issue [#1367](https://gitlab.earth.bsc.es/digital-twins/de_340-2/workflow/-/issues/1367))

Close #2471

----

 - [x] Changelog is updated.

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -14,6 +14,10 @@ def reader_instance(fesom_test_pi_original_2d_r200_fixFalse_reader):
 def data(fesom_test_pi_original_2d_r200_fixFalse_data):
     return fesom_test_pi_original_2d_r200_fixFalse_data
 
+@pytest.fixture(scope='module')
+def reader_ifs_tco79_long(ifs_tco79_long_reader):
+    return ifs_tco79_long_reader
+
 # aqua class for tests
 @pytest.mark.aqua
 class TestAqua:
@@ -108,12 +112,11 @@ class TestAqua:
         assert len(result.data_vars) == 0
 
 
-    def test_time_selection(self):
+    def test_time_selection(self, reader_ifs_tco79_long):
         """
         Test that time selection works correctly
         """
-        reader = Reader(model="IFS", exp="test-tco79", source="long",
-                        loglevel=loglevel)
+        reader = reader_ifs_tco79_long
         
         data = reader.retrieve(startdate='2020-03-01', enddate='2020-03-31')
         


jhardenberg:
One comment: we missed this problem because we actually have so good as no tests for CLIs, in particular the DROP one...

oloapinivad:
Thanks for extending DROP CLI! I feel this has been tested only via configuration files, and it is true that we are missing test for this.

However, I feel something redundant has been made concerning the data selection. I do not feel we need to extend the feature for `startdate` and `enddate`, as we already have this feature as far as I know, see L393.  

https://github.com/DestinE-Climate-DT/AQUA/blob/d6d0f3ea3eb6918fb8fdf7182e53665287faf80c/src/aqua/reader/reader.py#L387-L394

jhardenberg:
hmmm...I added it because it seemed to ignore it, let me check, thanks

src/aqua/reader/reader.py
```diff
@@ -390,7 +390,7 @@ def retrieve(self, var=None, level=None,
             if data is None or len(data.data_vars) == 0:
                 self.logger.error(f"Retrieved empty dataset for {var=}. First, check its existence in the data catalog.")
 
-            if startdate and enddate and not ffdb:  # do not select if data come from FDB (already done)
+            if (startdate or enddate) and not ffdb:  # do not select if data come from FDB (already done)
```
oloapinivad: well spotted!

PR #2496, opened 2025-11-20
Timeout in test job action

### LINKED ISSUE #2491: Timer in Gihub action for hanging tests to retrig them
Sometimes it happens (especially in local) that tests hangs and stop working (don't know why yet).  

Since the tests do not hang always, can all of you who experience the problem report the step at which the hanging appeared?
In this way, we could do an empirical survey through time to see if there's a pattern with common tests that hang, or if this is a more random behaviour.

@silviacaprioli @oloapinivad @mcadau @mnurisso @jhardenberg @maqsoodrajput @ghossh

tovaz92:
Enhance test job action with timeout and max retry logic.

Close #2491

----

 - [x] Changelog is updated.


tovaz92:
Since the tests do not hang always, can all of you who experience the problem report the step at which the hanging appeared?  
In this way we could see if there is a pattern related to certain tests triggering the hang, or wether this is more a random behaviour.

tovaz92:
Some time has passed since this PR was opened. The need for retry logic likely came from occasional random code hangs. I've already included a `--max-worker-restart=3` option, which restarts crashed workers (though this is a `pytest-xdist` feature, not a GitHub Actions retry mechanism).
In general, implementing `n-retries` to work around hanging code is not the best approach imho. Instead, it's could be better to detect when an action hangs and report these incidents in the related issue: https://github.com/DestinE-Climate-DT/AQUA/issues/2491.
Therefore, I would close this PR.

PR #2287, opened 2025-09-30
Developing AQUA in pypi

oloapinivad:
## PR description:

This PR explores how to publish AQUA in a way that is fully usable on pypi. 

- [x] Build the package
- [x] Publish in  testpypi (temporary name of `aqua-climate`)
- [x] Establish a workflow pypi pipeline that uses the testpypi version
- [ ] Run successfully tests with the testpy version
- [ ] Add a workflow to publish each new release on pypi
- [ ] Make a new release to reach pypi

## Issues closed by this pull request:

Close #1829 

## People involved:

@mnurisso @jhardenberg 

----

 - [ ] Documentation is included if a new feature is included.
 - [ ] Changelog is updated.

PR #2500, opened 2025-11-21
Going to PyPi

oloapinivad:
This seems very promising, we have been able to publish on testpypi and verify it works

https://test.pypi.org/project/aqua-core/

`builder.sh` is only to keep track of required passages

See this https://github.com/DestinE-Climate-DT/AQUA/pull/2287 for further details

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 ## [Unreleased]
 
 Unreleased in the current development version (target v0.21.0):
+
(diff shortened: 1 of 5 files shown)


.github/workflows/aqua-test-reusable.yml
```diff
@@ -145,13 +151,37 @@ jobs:
...
+            eccodes==2.41.0
+            cartopy_offlinedata
+            pandas
+            pip
+            hdf5
+            intake-esm
+            imagemagick
+      
       # ========================================
       # 4. LINTING
```
mcadau: 4a?

PR #2502, opened 2025-11-21
Update lumi-install

jhardenberg:
## PR description:

Update lumi install in a version which installs ONLY the framework

----

 - [x] Changelog is updated.


jhardenberg:
This is harmless, if somebody approves we can merge @tovaz92

tovaz92 (commented):
Is there a dupliaction of the question in lumi_install in lines 195-198?

tovaz92 (approved):
Looking good, approving

PR #2507, opened 2025-11-24
Fix aqua-diagnostics installation and console refactoring

oloapinivad:
## PR description:

This provides an initial fix to have a complete installation of AQUA-diagnostics, with the complete copying of files as we would like to have. I want to redesign a couple of features, and this does not support yet editable installation.


mcadau:
I agree with Matteo and I like the idea too! And this is supposed to be implemented in this same PR or in the near future?

mcadau (approved):

PR #2512, opened 2025-11-25
Console support for `--core` and `--diagnostics`flags

oloapinivad:
## PR description:

This PR further rework the console, and builds on `fix-installation` branch to be merged yet. 

It will allow for flexible installation of aqua core and diagnostics, following this schema: 

```bash
# Default installation, as currently deployed (core + diagnostics)
aqua install my-machine

# Installation of the core
aqua install my-machine --core

# Installation of the diagnostics only - will fail if core is not available
aqua install my-machine --diagnostics

# Standard installation (as no flags provided)
aqua install my-machine --core --diagnostics

# Core editable from custom path custom, diagnostics not installed
aqua install my-machine --core /path/to/aqua/repo

# Diagnostics editable from custom path, core not installed - will fail if not there
aqua install my-machine --diagnostics /path/to/diagnostics/repo

# Core standard, diagnostics editable
aqua install my-machine --core --diagnostics /path/to/diag

# Core editable, diagnostics standard
aqua install my-machine --core /path/to/core --diagnostics

# Both editable 
aqua install my-machine --core /path/to/core --diagnostics /path/to/diag
```

Still need to be tested completely, also because it is not possible to develop tests for diagnostics in this repo

### DIFF
--- a/aqua/core/console/install.py
+++ b/aqua/core/console/install.py
@@ -253,7 +253,7 @@ class InstallMixin:
                     self.logger.error('Cannot install diagnostics without core. Install core first or use full installation.')
                     sys.exit(1)
                 if install_info['diagnostics']['installed']:
-                    self.logger.error('Diagnostics component is already installed. We cannot add it again, please use \'aqua install\'')
+                    self.logger.error('Diagnostics component is already installed. We cannot add it again, please use \'aqua uninstall\' before')
                     sys.exit(1)
                 self.logger.info('Core already installed (%s mode), adding diagnostics component',
                            install_info['core']['mode'])


aqua/core/console/install.py
```diff
@@ -46,16 +264,67 @@ def install(self, args):
...
-    
+
         # define the template path
         self.templatepath = os.path.join(self.configpath, 'templates')
+    
+        # Check current installation status
+        install_info = self._check(silent=True, return_info=True)
+        
+        # Determine installation mode for each component
+        selective_install = args.core is not None or args.diagnostics is not None
```
mnurisso: it is not super clear how this is used. Is a real `or` or it should be more a xor? Do you want it true as well when both flags are there?

oloapinivad: yes, if both are specified both are installed, if none is specified both are installed, if only one is specified only one is installed. I have just made a simplification, pushing now.

jhardenberg:
In fact, what happens if the user switches position of the arguments and instead of 

```
aqua install my-machine --core
```

we write
```
aqua install --core  my-machine 
```
Will `--core` eat up `my-machine`?

To avoid ambiguity,  an alternative would be:
```
# Default installation, as currently deployed (core + diagnostics)
aqua install my-machine

# Installation of the core
aqua install my-machine --core

# Installation of the diagnostics only - will fail if core is not available
aqua install my-machine --diagnostics

# Standard installation (as no flags provided)
aqua install my-machine --core --diagnostics

# Core editable from custom path custom, diagnostics not installed
aqua install my-machine --core-path /path/to/aqua/repo

# Diagnostics editable from custom path, core not installed - will fail if not there
aqua install my-machine --diagnostics-path /path/to/diagnostics/repo

# Core standard, diagnostics editable
aqua install my-machine --core --diagnostics-path /path/to/diag

# Core editable, diagnostics standard
aqua install my-machine --core /path/to/core-path --diagnostics

# Both editable 
aqua install my-machine --core-path /path/to/core --diagnostics-path /path/to/diag
```

oloapinivad:
> What is the magic which makes both `aqua install my-machine --core /path/to/core --diagnostics /path/to/diag` and `aqua install my-machine --core --diagnostics` possible? So `--core` and `--diagnostics` are both flags (install core or diagnostic yes/no) and keyword arguments providing the paths? I do not dislike it but I fear that it is not very standard and possibly confusing ....

Yes, this is the feature I asked about a few days ago :-) `--core` install the default version, and `--core path` install the editable way. It avoids the creation of 4 keys to be passed to match all the options

Which is exactly what you suggest in your comment below 😄 and it is what I tried to avoid with the current setup :-)

mnurisso:
> In fact, what happens if the user switches position of the arguments and instead of
> 
> ```
> aqua install my-machine --core
> ```
> 
> we write
> 
> ```
> aqua install --core  my-machine 
> ```

I see quite often a positionally mandatory argument before the cli args. If you to the second command it will fail because there is no machine name if that part of behavior didn't change

PR #2526, opened 2025-11-28
Address parallel failure due to `nemo-curvilinear.yaml`

oloapinivad:
## PR description:

We are still facing random failure of tests due to access of the `nemo-curvilinear.yaml`.  See here: https://github.com/DestinE-Climate-DT/AQUA/actions/runs/19742925440/job/56570971328 

Claude suggests that this is related to the absence of the `loadgroup` call into our refactored workflow, and tests are not accessed in parallel.
 
However, I suspect this is not enough. The grid files are modified by the grid builder test, and this create an extra file that might be accessed in parallel by other tests loading the grids configuration files.

### DIFF
--- a/.github/ISSUE_TEMPLATE/01-bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/01-bug-report.yml
@@ -50,18 +50,17 @@ body:
       description: Which AQUA version are you using?
       options:
       - main
+      - v0.21.0
       - v0.20.1
       - v0.20.0
+      - v0.19.5+op
+      - v0.19.5
       - v0.19.4+op
       - v0.19.4
       - v0.19.3+op
       - v0.19.3
-      - v0.19.2+op
-      - v0.19.2
-      - v0.19.1+op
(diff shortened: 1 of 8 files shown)


oloapinivad:
I tried to put together the regrid and the grid builder but now it fails in the catalog... 😕 
@tovaz92 any clue what is going on?

tests/utils_tests/cleanup.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    Registry for tracking and cleaning up files created by tests.
+    
+    This class provides a simple way to clean up test artifacts
+    that are created in the AQUA configuration directory (~/.aqua).
+    """
+    
+    # Grid files created by tests that need cleanup
+    # Paths are relative to configdir (e.g., ~/.aqua/grids/regular.yaml)
+    FILES_TO_CLEAN = [
```
oloapinivad: perhaps you can move this files to the top of the files, so that it is simpler to see them if they need to me modified

PR #2535, opened 2025-12-01
Fixes AQUA paths of grids-downloader

jhardenberg:
## PR description:

Small fix to paths due to new structure

----

 - [ ] Changelog is updated.


jhardenberg:
coming back to this one: as long as the grids downloader is still with us we might as well fix it (not to keep a broken piece of software around). So maybe transform this into a PR to encourage use of DVC instead, but in the meantime merge this? @oloapinivad

PR #2545, opened 2025-12-05
Fix `GridBuilder` with healpix 3d grid

oloapinivad:
## PR description:

We are experiencing failures in GridBuilder when running on 3d healpix grids. This has been previously reported also by @maqsoodrajput when working with EERIE data (I think). 

There were two problems:
- due to a bug in ClimateDT data, the data model is not working as it should and vertical dimension is not identified. `GridBuilder` has the option to force the `vert_coord`, but this was not exposed in the command line
- There was a dumb typo in the function that was extracting the metadata from the healpix grids, i.e. the one that computes which CDO grid should be used. It was reading from 3d data instead from 2d, meaning that the healpix detection was wrong (too large grid)

Both issues are fixed now, and should work smoothly for operations. I added a few debug and another test for 2d oceanic data. Cannot test healpix 3d data since generation is too slow, but offline it works. 

----
 - [x] Changelog is updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Unreleased in the current development version (target v0.22.0):
 
 AQUA core complete list:
 - Fix for 3d grid generation of `GridBuilder` (#2545)
+- Fix failing console test (#2549)
 - New `filter_key` option to filter NetCDF files before loading them based on year (#2543)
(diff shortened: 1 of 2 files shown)


oloapinivad:
This works, I also added a more complete console test. I think we need to verify the functionality of the tool on real data generation to be fully sure it work.

PR #2574, opened 2025-12-16
Tiggi/fix aqua web push 2

tiggi:
Actually fix the push_analysis.sh script instead of the almost-fix previously committed.

A trap rm -rf with a relative path inside a script that changes directories isn't a good idea.
Fix that by feeding the trap an absolute path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 Unreleased in the current development version (target v1.0.0):
 
 Complete list:
+- Aqua web push folder management improvements (#2574)
 - Improve fldstat handling of gaussian grids (#2584)
 - Data model now uses a ranking system to identify coordinates (#2585) 
 


cli/aqua-web/push_analysis.sh
```diff
@@ -349,6 +349,7 @@ else  # Otherwise, use the second argument as the experiment folder
 fi
 
 if [ $update -eq 1 ]; then
+    git pull
```
jhardenberg: will this not fail if some other instance modified updated.txt upstream?
Maybe better not to do this here and just add `--force` to the following `git push` ?

cli/aqua-web/push_analysis.sh
```diff
@@ -285,8 +285,7 @@ if [ $localrepo -eq 1 ]; then
 else
     log_message INFO "Clone aqua-web from $repository"
     repo=aqua-web$$
-    absoluterepo=$(realpath $repo)
-    trap "rm -rf $absoluterepo" EXIT
+    trap "rm -rf '$PWD/$repo'" EXIT
```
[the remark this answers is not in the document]
tiggi: The trap is incorrect as it is now, so don't merge this version.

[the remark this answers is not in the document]
tiggi: Having tested this, realpath works on nonexisting directories, so the code works as intended.

realpath foo generates $PWD/foo as output.

cli/aqua-web/push_analysis.sh
tiggi: Changed to use mktemp instead, so that it's always a truly unique directory and always an absolute path.
aqua-web$$ is a temporally unique value between 1 and 4 million, and it wraps quite quickly on the autosubmit VM.

tiggi: the problem is that you do "git clone, build/upload images for 15 minutes, update updated.txt, commit + push". If any other parallell process manages to push during this time, you get a conflict. If you insert the git pull before you update
updated.txt the window of conflict shrinks to a couple of seconds and it's much less likely to conflict, since the repo is suddently up to date with the server.

This is not a teoretical issue, I have to clean up the messes that this causes a lot. Reading logs led me to this script and it's obviously not written with paralellism in mind in any way.

PR #2587, opened 2025-12-20
Avoid race conditions to grids yaml files

tovaz92:
Avoid racing condition and fix possible issues with tests related to `nemo-curvilinear.yaml`.

## People involved:
@oloapinivad 

----

 - [x] Changelog is updated.


mnurisso:
This may be a similar issue:

https://github.com/DestinE-Climate-DT/AQUA/actions/runs/20455845095

PR #2600, opened 2025-12-30
Optimize `_filter_kwargs` in Reader

### LINKED ISSUE #2404: Disentangle kwargs, chunk and engine treatment in Reader
### Feature request

In the current implementation of the Reader we're using a method called `_filter_kwargs` to manage kwargs, chunk (which is an argument) and engine (which is a kwarg but only for FDB sources).
This is not efficient, because the method is called also when only chunks or engine are present but no other kwarg is there.
This issue is to discuss a refactor of the method, while now it is not high priority to change it.

tovaz92:
Refactor of `_filter_kwargs` increasing its efficiency if `kwargs` is empty dictionary. Reduce overhead when initializing `Reader` instances with only `chunks` or `engine` parameters, avoiding unnecessary `catalog` inspection.

Add lazy loading for `intake_user_parameters`.  Also skip inefficient catalog `describe()` calls when no user `kwargs` is provided.

Note: Potentially redundant manual default value handling was removed, relying on intake-esm's built-in merging.

Close #2404

----

 - [x] Changelog is updated

### DIFF
--- a/aqua/core/reader/reader.py
+++ b/aqua/core/reader/reader.py
@@ -671,7 +671,7 @@ class Reader():
             self.logger.warning('Keeping chunks=%s in the filtered kwargs', self.chunks)
             filtered_kwargs['chunks'] = self.chunks
 
-        # Check for missing required parameters and apply defaults with logging
+        # Check for missing required parameters and apply defaults, with logging
         # We identify parameters that are valid but not present in either filtered_kwargs or intake_vars
         
         # params that are already covered by user kwargs or machine-specific intake_vars


aqua/core/reader/reader.py
```diff
@@ -624,54 +621,48 @@ def _format_realization_reader_kwargs(self, kwargs: dict):
...
+        if kwargs:
+            params = [elem.get('name') for elem in self.intake_user_parameters]
+            
+            filtered_kwargs = {k: v for k, v in kwargs.items() if k in params}
+            
+            unfiltered_kwargs = set(kwargs) - set(filtered_kwargs)
+            for key in unfiltered_kwargs:
+                self.logger.warning('kwarg %s is not in the intake parameters of the source, removing it', key)
+                
+            # Note: intake-esm automatically applies catalog defaults for missing parameters when creating the source
```
mnurisso: Here it's a bit a pity to lose logging, but if efficiency is higher it is fine

PR #2606, opened 2026-01-09
Add HighResMIP grids config

tovaz92:
As we want to use HighResMIP data, we need the config grids entries from the generated grids.

----

 - [ ] Changelog is updated.


jhardenberg:
#2627 introduced 9 grids, here not all those models are represented and there is one more (AWI). Should they be homogenized?

PR #2627, opened 2026-01-21
HighResMIP atm grids

mnurisso:
## PR description:

As per title

 - [x] Changelog is updated.

### DIFF
--- a/.github/ISSUE_TEMPLATE/01-bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/01-bug-report.yml
@@ -51,11 +51,14 @@ body:
       description: Which AQUA version are you using?
       options:
       - main
+      - v1.0.0a3
       - v1.0.0a2
       - v1.0.0a1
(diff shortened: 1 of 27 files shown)


aqua/core/config/grids/EC-EARTH3.yaml
```diff
@@ -0,0 +1,6 @@
+grids:
```
oloapinivad: why EC-EARTH3 all upper case? the name of the model is EC-Earth3, or given that all the other model are lower case you cn keep it as the others.

oloapinivad (approved):

oloapinivad:
Only doubt, does it make sense to have a single file for any of them or it is better to have them together in a single highresmip file?

tovaz92:
So, do you mean having paths like: `'{{ grids }}/HighResMIP/<model>/<filename.nc>'`?

tovaz92 (approved):

PR #2649, opened 2026-01-27
Add cleanup of partial catalog installations with `aqua add`

### LINKED ISSUE #2616: aqua add in editable mode fails if it has failed before
### machine

Others

### catalog

climatedt-o25.1

### version

main

### What happened?

Trying to add catalogues in editable mode on mafalda  I was confused for a moment if I had to point as a path to 
`/work/users/jost/aqua/Climate-DT-catalog` or to `/work/users/jost/aqua/Climate-DT-catalog/catalogs/climatedt-o25.1 `, first trying with the first option and I did 
````
aqua add -e /work/users/jost/aqua/Climate-DT-catalog climatedt-o25.1
````
AQUA then answers with the quite obscure message (that is also a problem) "2026-01-15 16:40:11 :: AQUA :: ERROR    -> No AQUA installation found!".

Anyway, realizing my error, if afterwards I try the correct
````
aqua add -e /work/users/jost/aqua/Climate-DT-catalog/catalogs/climatedt-o25.1 climatedt-o25.1
````

I still get the same error.
I discovered that the problem is that the first wrong command already added the entry
````
catalog: climatedt-o25.1
````
to `.aqua/config-aqua.yaml` and the second command file fails if it already exists.

Solution: 
1) make sure that if the command fails no catalog entries have been added to the config file.
2) Provide a better error message

```python

```

### Error messages

_No response_

None

tovaz92:
Improves catalog installation for editable mode and improve error message.

Close #2616

----

 - [x] Changelog is updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Unreleased in the current development version (target v1.0.0):
 
 Complete list:
 - Add cleanup on failure for console `aqua add` catalog addition (#2649)
+- Console: Single API call for `aqua add` (#2675)
 - `plot_single_map()` better Healpix handling (#2671)
 - Reader: ignore regridded flag if not usable (#2664)
 - Smmregrid weights are corrected by fixer/datamodel (#2639)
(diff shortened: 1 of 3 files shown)


jhardenberg:
@tovaz92 can you explain what these chanegs actually do ?

PR #2656, opened 2026-01-30
[GSV] Remove necessity of FDB paths if engine is polytope

### LINKED ISSUE #2590: Remove FDB_home_bridge necessity for polytope only data
### Feature request

I'm testing nextgems4 databridge access on my local machine.
Maybe a fdb_home_bridge is available on lumi, however, as a user who doesn't know it, I need anyway to provide a path (obviously fdb_home_bridge: pippo) otherwise the code will not work. We can handle better this with a warning, but I can normally access data without the metadata.

### LINKED ISSUE #2618: fdb_home is still needed when using polytope engine for data fully on the bridge
### machine

Others

### catalog

climatedt-o25.1

### version

main

### What happened?

Trying to read data which are now residing entirely on the bridge (e.g. historical-1990 IFS-NEMO o25.1) using polytope, the catalogue still needs to have `fdb_home` line, even if this then is not used and its value dos not make any sense anymore.

```python

```

### Error messages

_No response_

None

silviacaprioli:
## PR description:

Using polytope, the catalogue still needs to have `fdb_home` and `fdb_info_file` lines, even if they are not used. This PR fixes this issue

Close #2590
Close #2618

 - [ ] Changelog is updated.
 - [ ] environment.yml and pyproject.toml are updated if needed, together with the lumi installation tool. Installation instructions are updated if new conda dependencies are added.


aqua/core/gsv/intake_gsv.py
```diff
@@ -80,7 +80,7 @@ def __init__(self, request, data_start_date, data_end_date, bridge_start_date=No
         """
 
         self.logger = log_configure(log_level=loglevel, log_name='GSVSource')
-        self.engine = engine
+        self.engine = engine if engine else 'fdb'
```
jhardenberg: Why is this needed? Why not leave `engine='fdb' `above and `self.engine = sengine` here??  I fail to see the subtle difference

silviacaprioli:
```
def __init__(self, ..., engine=None, ...):

    self.engine = engine or 'fdb'
    self.dummy_run = (engine is None)
```

Not sure anymore this logic is correct

silviacaprioli:
> For this to work we need to always pass explicitly an engine argument to the driver ... maybe this is not the case? Does the reader always pass engine=something? Do the tests?

The reader has `engine='fdb'` as default, no problem with that.  Perhaps passing explicitly the engine in the tests is the way to go

PR #2693, opened 2026-02-16
Missing attributes after regridding

### LINKED ISSUE #2692: Regridding loses attributes
### machine

Levante

### catalog

obs

### version

main

### What happened?

After regridding a Dataset (not a dataarray) we lose attributes, so that the output is for example missing attribute `AQUA_catalog`. This leads also to diagnostic error [#149](https://github.com/DestinE-Climate-DT/AQUA-diagnostics/issues/149)

```python
from aqua import Reader
reader = Reader(catalog='obs', model = 'BERKELEY-EARTH', exp='aqua-filled', source='r100-monthly', regrid='r100')
data = reader.retrieve(var='2t')
data.aqua.regrid()['2t']
```

### Error messages

_No response_

None

jhardenberg:
## PR description:

The problem is probably in smmregrid, but this at least adds a test for the problem of lost attributes described in #2692

Close #2692

----

 - [x] Changelog is updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 Unreleased in the current development version (target v1.0.0):
 
 Complete list:
+- Add a test for attributes after regridding (#2693)
 - Regridding of datasets to ignore bounds (#2678)
 - Fix units for FESOM thetao variable in fixer (#2686)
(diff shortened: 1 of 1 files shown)


jhardenberg:
Ok, the updated smmregrid solves the problem . So as soon that is in its main, we can merge this one too.

oloapinivad (approved):
Fine with this. I am working on a smmregrid release anyway, we can change the pin in a second moment

PR #2700, opened 2026-02-18
Automatically initialize regridders with loaded weights in the Regridder class

### LINKED ISSUE #2837: Regridder documentation doubt
I was using the Regridder alone with this call:

```
regridder = Regridder(cfg_regrid, data=data, loglevel=args.loglevel, src_grid_name=src_grid_name)
regridder.weights(tgt_grid_name=resolution, regrid_method="con", reader_kwargs=reader_kwargs)
data_r = regridder.regrid(data)
```

I had to modify to:

```
regridder = Regridder(cfg_regrid, data=data, loglevel=args.loglevel, src_grid_name=src_grid_name)
weights = regridder.weights(tgt_grid_name=resolution, regrid_method="con", reader_kwargs=reader_kwargs)
regridder.initialize(weights)
data_r = regridder.regrid(data)
```

This behavior is not in the documentation. Do we need to update it?

mnurisso:
## PR description:

I tested a Regridder alone workflow such as:

```python
regridder = Regridder(cfg_regrid, data=data, loglevel=args.loglevel, src_grid_name=src_grid_name)
regridder.weights(tgt_grid_name=resolution, regrid_method="con", reader_kwargs=reader_kwargs)
data_r = regridder.regrid(data)
```
In the current implementation it gives error:

```python
2026-02-18 16:23:22 :: Regridder :: ERROR    -> Regridder for vertical coordinate 2d not found.
2026-02-18 16:23:22 :: Regridder :: ERROR    -> Cannot regrid variable tprate
```

This because the self.initialize method is not used. I can add it manually in my code but this is exactly the example from the documentation so I think a fix should be inserted in the code and this is what I am proposing

## People involved:

@oloapinivad 

 - [x] Changelog is updated.
 - [x] Added test


mnurisso:
Hi @jhardenberg could you check if you find any problem in this additional step?
I had problem running a script in which the `Regridder` class alone had to generate weights and immediately use them because they were not initialized. I think in the `Reader` the initialization is handled somewhere else so I'm not sure if I'm adding an overhead.

mnurisso:
> The reason why the initialization of theregridders was delayed in the `initialize()` method is to allow the application of the datamodel and of the fixers on the weights, see lines 322-326 of the reader. So in the reader the weights generation is called, the weights are fixed and only then the initialization of the regridders occurs. As you notice, this breaks the usage of the Regridder class alone.
> 
> The example should be fixed anyway, because `regridder.weights()` norw actually returns the weights.
> 
> We could either just change the docs saying that one needs to call initialize after computing the weights (but I admit that for a user this looks complicated) or introduce a 'double' behaviour of the `weights()` method by adding a `initialize` key as an argument which will be True by default and which will be used with False by the reader instead. This way we could preserve the behaviour described in the docs and at the meantime leave the reader the opportunity to apply the fixer/datamodel .....

Maybe this can be discussed on monday if there is extra time, I have no problem in calling explicitly the initialize method but indeed I'm not a beginner user and the documentation was wrong about this specific case.

jhardenberg:
Ok, I implemented our plan above and added a test which without the fix would fail

jhardenberg:
With this docs do not need to be updated

oloapinivad (approved):

jhardenberg:
@mnurisso so, should we merge?

oloapinivad:
changelog is missing

oloapinivad:
BTW, with this shall we close and not merge #2864?

mnurisso:
If we merge this yes, I'd close #2864 
The PR looks nice to me now

silviacaprioli:
Changelog is missing. I'd merge it, then.

mnurisso:
Merging if tests are passing

PR #2702, opened 2026-02-18
[Lint] reduce warnings

tovaz92:
As per title, reducing some tedious warnings that appears in the log files of the tests.

 - [x] Changelog is updated.

### DIFF
--- a/aqua/core/reader/reader.py
+++ b/aqua/core/reader/reader.py
@@ -8,7 +8,6 @@ import re
 import intake_xarray
 import xarray as xr
 import pandas as pd
-import numpy as np
 from metpy.units import units
 
 from smmregrid import GridInspector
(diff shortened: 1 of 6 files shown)


oloapinivad:
If this is associated to the flake8 linting warning, I would assume that using black or autopep8 could be the way to go

jhardenberg:
Ok, if this is ony about linting warning  I would change the title to 'reduce linting warnings" or similar, for clarity. I was assuming you were talking about logger messages at the warning level.

If this is about linting (a thing I always thought we should do), then we should do as suggested by  @oloapinivad  : a massive black or autopep8  run on ALL of our codebase at once, not only selected pieces.

There is a problem though: anything like that, once merged into the main will immediately make all file changes of all current PRs unreadable. So I would suggest to do this only when we have very few PR left .....

oloapinivad:
I think this one - following the ruff implementation - will be deprecated, correct?

PR #2703, opened 2026-02-19
Test  chunking ci

jhardenberg:
## PR description:

Just to test if adding chunking at time: 24 for the test-tco79.yaml long sources solves the test hang issues addressed also in #2701

The ci/cd action is modified so that it uses the new `chunk_ci_long` branch of Climate-DT-catalog (which adds a chunking at 24h to all 'long' sources in `test-tco79.yaml`
Not to be merged

### DIFF
--- a/.github/workflows/aqua-test-reusable.yml
+++ b/.github/workflows/aqua-test-reusable.yml
@@ -297,8 +297,12 @@ jobs:
       # ========================================
       # 10. RUN TESTS
       # ========================================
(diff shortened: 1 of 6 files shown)


oloapinivad:
Unfortunately it tells me "tests started five hours ago" so I guess it is not the solution

jhardenberg:
Ok, opened https://github.com/DestinE-Climate-DT/Climate-DT-catalog/pull/234 and closing this

PR #2721, opened 2026-02-24
V0.13 ubuntu container 24.04

mnurisso:
## PR description:

Updating the container to ubuntu 24.04 and adapting the dependencies to a new release
cc @TracyMcBean 

Short list of changes:
- container base is ubuntu24:04
- intake-esm and fdb auto date tests are disabled (not operational features)
- pin most of dependencies to what is in the v0.13.9 container
- explicit installation of setuptools to support `pkg_resources` necessity (now deprecated)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ Unreleased in the current development version:
 ## [v0.13.10]
 
 Hotfixes:
+- Explicit pin of main dependencies to avoid breaking changes (#2721)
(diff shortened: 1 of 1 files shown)


jhardenberg:
is removing intake-esm still needed after pinning everything?

mnurisso:
> is removing intake-esm still needed after pinning everything?

idk, is it an important feature? This will be used only to finish the icon run

PR #2803, opened 2026-04-03
[DROP] Further check to verify that internally concatenated files are complete

### LINKED ISSUE #2668: [Drop] Safety check before concatenation of yearly files
### Feature request

It would be useful to implement a further safety check in DROP so that before files are concatenated into yearly blocks the single monthly file is scanned and verified, since we had experience of creation of yearly files with missing month.

tovaz92:
Add a more fine-grained check during file concatenation in DROP.  
Verify that all 12 monthly files exist, are complete, and contain time coordinates matching the expected year and month before any concatenation takes place.  

Close #2668

----

 - [x] Changelog is updated.


oloapinivad:
This will create massive conflicts with #2824, we need to decide how to proceed

tovaz92:
This do not provide substantial changes that can be used in the newer architecture in #2824, thus I will close this.

PR #2818, opened 2026-04-16
Experimental removal of fdb from main dependencies

### LINKED ISSUE #2823: Allow FDB dependencies to be optional packages
### Feature request

We want to be able to install AQUA without FDB because of Conda deploy

oloapinivad:
## PR description:

We are exploring if AQUA core can be shipped without FDB dependencies as default, which might potentially allow the publication of aqua on conda-forge. A new test is introduced to verify if without it everything is working. Question is if we want those tests to run frequently or not.

 - [ ] Changelog is updated.
 - [ ] environment.yml and pyproject.toml are updated if needed, together with the lumi installation tool. Installation instructions are updated if new conda dependencies are added.

### DIFF
--- a/docs/sphinx/source/installation.rst
+++ b/docs/sphinx/source/installation.rst
@@ -34,6 +34,9 @@ This can be achieved with:
 The same environment is available in the AQUA-core GitHub repository in the ``environment-pypi.yml`` file.
 
 .. note ::
+    If you want to access ClimateDT data, you will need to run `pip install aqua-core[fdb]`
+
+.. note ::
     If you need to access data written in a local FDB database (not polytope), you need to install the FDB5 library.
     The FDB5 library is not available in the conda-forge repository, so you need to install it manually.
     If you are working on a supported HPC, you can check the corresponding section for more information in the :ref:`HPC installation <installation-hpc>` section.
@@ -56,6 +59,8 @@ You can install them with the following command:
     pip install aqua-core[docs]
     pip install aqua-core[notebooks]
     pip install aqua-core[tests]
+    pip install aqua-core[style]
+    pip install aqua-core[fdb]
 
 Or to install all the extra dependencies:
 
@@ -115,8 +120,6 @@ At this point, you should have successfully installed the AQUA package and its d
 .. note ::
 
     By default, the environment file installs the cloned version of AQUA in editable mode with ``pip install -e .[all]``.
-    For development workflows, switch that line in ``environment.yml`` to ``pip install -e .[precommit,all]`` and then
-    run ``pre-commit install`` once in the repository.
 
 Update of the environment
 ^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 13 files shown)


oloapinivad:
@jhardenberg if you check the tests, they are failing not because the gsv is not there but rather because the fdb home is not found. how is that possible? at the beginning of the intake gsv module we have 

```python
# Test if FDB5 binary library is available
try:
    from gsv.retriever import GSVRetriever

    gsv_available = True
except RuntimeError:
    gsv_available = False
    gsv_error_cause = "FDB5 binary library not present on system or outdated"
except KeyError:
    gsv_available = False
    gsv_error_cause = "Environment variables for gsv, such as GRID_DEFINITION_PATH, not set."
```

that should raise a specific error...

jhardenberg:
apparently the gsv retriever  was called with `engine='fdb',` so that  `self.dummy_run=false` (so that fdbhome is checked). Maybe we should simply set always `self.dummy_run=false`  if  `gsv_available = False`....

jhardenberg:
... on the other hand we have
```
        if not gsv_available:
            raise ImportError(gsv_error_cause)
```
so it is unclear how the code could reach the stage where this bacame a problem ...

oloapinivad:
Ah! this now works, ready for review! @mnurisso @jhardenberg 
I added a `polytope` mark and split the fdb packages as optional. Of course, you still need gsv retriever to access climatedt data unless we write a polytope intake reader

docs/sphinx/source/installation.rst
```diff
@@ -33,6 +33,9 @@ This can be achieved with:
 
 The same environment is available in the AQUA-core GitHub repository in the ``environment-pypi.yml`` file.
 
+.. note ::
```
mnurisso, line 36: we need to remember to update the user documentation also in the community resources repo when this version will be released on pypi

environment.yml
```diff
@@ -28,7 +28,7 @@ dependencies:
   - gcsfs<=2026.2.0
   - healpy<=1.18.1
   - intake==0.7.0
-  - intake-xarray<=2.0.0
+  - intake-xarray<=0.7.0
```
mnurisso, line 31: I don't understand this change here

mnurisso (approved):
I think this can be merged, I also think we need a new release as soon as possible so that these pypi qol improvements will be active

environment.yml
oloapinivad: it turns out that `intake-xarray==2.0.0` does not work, I cannot recall exactly but there in the previous version we were installing an old version of intake, which does not happen with conda

docs/sphinx/source/installation.rst
oloapinivad: that's true, shall we open an issue?

mnurisso: Yes, I'll do it now

PR #2820, opened 2026-04-21
GitHub agent.md

### LINKED ISSUE #2817: Investigate the integration of `AGENTS.md` or `CLAUDE.md`
### Feature request

@mcadau will check what is the best practice to integrate a AI agent instruction

oloapinivad:
## PR description:

I have asked Claude to create a `agent.md` to be used by copilot. I told him to scan my code and follow some directives (code reuse, minimal intervention, etc) to generate the file. I am honestly impressed by the result. It would be great if anyone here review the file so that we can merge this and test if it is effective. I am starting to wonder if even more complex tasks as the intake migration can be done with claude. 

Close #2817 

 - [x] Changelog is updated.


mnurisso:
Possible suggestions (also chatgpt based):

The high level architecture we like is too implicit for an LLM, we may want to have something more clear like:

```
## AQUA Processing Pipeline (Mental Model)

Data flow is always:
Catalog → Reader → Fixer → DataModel → Regridder → Statistics → Output
- Reader orchestrates everything
- Each step must remain independent and composable
- No step should bypass or duplicate another

Reader is the ONLY entry point for data workflows.
- Do not create alternative pipelines
- Do not bypass Reader to call Fixer/DataModel/Regridder directly in user-facing code
```

Same for intake, were we may want to emphasize the idea:

```
## Intake Usage Rules
- All data access MUST go through Intake catalogs unless explicitly justified
- Do not hardcode file access patterns already representable in catalogs
- Extend Intake drivers when supporting new formats (e.g. FDB), do not bypass Intake
- Use syntax from intake 0.7.0

## Adding a new data format
1. Extend Intake driver
2. Register in catalog YAML
3. Avoid modifying Reader unless necessary
```

Same for parallelism:

```
## Dask & HPC Guidelines
- Avoid triggering computation (`compute`, `persist`) inside library code
- Leave execution control to the user or workflow layer
- Chunking strategy should be defined at Reader level, not downstream
- Never assume a specific scheduler (threads/processes/distributed)
```

We mention the Error handling, saying where to find them, but not how to use them (maybe this is our fault and it is not fully clear to us as well)

We should add what to follow, but also what to avoid:

```
## Forbidden Patterns
- Calling `.compute()` inside core modules
- Converting to numpy (`.values`) unless strictly necessary
- Looping over xarray dimensions instead of vectorized ops
- Re-reading the same dataset multiple times
- Creating new xarray datasets when transformation can be in-place
```

We should state explicitely to not breack backward compatibility when introducing modifications.

We should change: Always explore aqua/core/ before writing new code into something more clear:
```
Before implementing:
1. Search for existing class handling similar logic
2. Extend it if possible
3. Only create new module if no extension point exists
```

A suggestion is to reduce the map table to be sure it does not become something to maintain (for example for APIs changes)

mnurisso:
Final remarks:

- AGENTS.md (not as hidden file) would be required to be more effective, or it may be avoided by agents
- We may want to reference it in the README, like:

```
## 🤖 AI Contribution Guidelines

If you are an AI agent, please read AGENTS.md in the root directory
```

oloapinivad:
> * AGENTS.md (not as hidden file) would be required to be more effective, or it may be avoided by agents

Are you sure about this? My understanding is that .agents is read by github copilot, which what we are all using

oloapinivad:
I integrated the suggestions from Matteo and restructured it following sonnet suggestion, looks nice to me. we can try it and see if provides an improvement.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Unreleased in the current development version (target v1.0.0):
 Main changes:
 
 Complete list:
+- Adding LLM directives in `.agent.md` for Copilot (#2820)
```
mnurisso: change the filename in the changelog

AGENTS.md
```diff
@@ -0,0 +1,214 @@
...
+- Use `Reader(model=..., exp=..., source=...)` to load test data; avoid raw `xr.open_dataset` unless testing I/O directly
+- Prefer shared fixtures in `tests/conftest.py` with `session` or `module` scope to avoid repeated heavy retrievals
+- Favour `pytest.mark.parametrize` over duplicated test bodies
+- When testing write paths, verify that Dataset-level attributes (especially `history`) are present in the output file — attribute loss through slicing or DataArray→Dataset conversion is a common silent failure
+
+---
+
+## Integration Checklist
+
+- [ ] `xr.set_options(keep_attrs=True)` set at module level
```
mnurisso: This is now the default for xarray, so it is not an error but it is redundant

mnurisso (approved):
Very nice, added a couple of minor comments, then we can merge directly

oloapinivad:
Thanks, updated, merging!

PR #2824, opened 2026-04-21
Enabling zarr/icechunk writing in DROP

### LINKED ISSUE #1888: LRA really wants to go to zarr
### Feature request

One useful we could implement is to allow the LRA (which is slowly becoming our data extraction tool at multiple scales/resolution) to natively write in zarr. 

I did some tests and it seems quite straightforward, see the code snippet below:

```
import os
import xarray as xr
import shutil

def should_append(ds_existing, ds_new):
    try:
        existing_times = ds_existing.time.values
        new_times = ds_new.time.values

        if any(t in existing_times for t in new_times):
            print("Time overlap detected, skipping append.")
            return False
        return True
    except Exception:
        return True

def zarrwrite(data, path):
    """Write a variable to a zarr store."""
    if os.path.exists(path):
        try:
            ds = xr.open_zarr(path, consolidated=False)
        except Exception as e:
            print("Zarr store is corrupted or incomplete, removing it")
            shutil.rmtree(path)
        
    if os.path.exists(path):
        if should_append(ds, data):
            print("Appending data to existing Zarr store")
            data.to_zarr(path, mode='a', 
                         append_dim='time', consolidated=False)
    else:
        print("Zarr store does not exist, creating it")
        data.to_zarr(path, mode='w', consolidated=False)


var='2t'
path = "/Users/paolo/Desktop/Zarr/2t.zarr"
#shutil.rmtree(path, ignore_errors=True)
for month in range(1, 8):
    print(f"Processing month {month}")
    sub = data[var].sel(time=data.time.dt.month==month)
    fld = reader.timmean(sub, freq='monthly')
    zarrwrite(fld, path)
```

This already includes some basic safety check to avoid overwriting/infinite appending and verifies that AT least metadata are stored correctly (which should already prevent us for big failures). BTW this goes beyond the need of having yearly files as done in netcdf since we create a single object which should be able in principle to accomodate multiple years and variable. Zarr allows for parallel writing so in principle I do not foresee race conditions, but it is possible to set up lockfiles. 

Possible issues that I see:
1.  Consolidation, which should be done when the simulation is over and it might not be obvious how to control this when running the LRA by hand.
2. If a single chunk (one month, or another minimum time unit) fails during the writing, I suspect this will BREAK entirely the zarr structure. I do not know if there is a safety check which can be implemented, but it can be quite annoying.
3. Controlling chunking might not be trivial. If we work with monthly blocks, it will not be efficient once the timeseries get very large. 
4. There should be a way to control loseless compression but it seems xarray does not have the same flexbility as for netcdf encoding. 

I tag also @sebastianbeyer since we discussed this with him this morning, and he might have some suggestions/opinions since he already worked quite a lot in this direction.

### LINKED ISSUE #1920: Allow for disabling the catalog entry generator

### LINKED ISSUE #2334: Explore capabilities of Icechunk for LRA/Zarr archives
### Discussion

I noticed that icechunk has a structure which is based on a sort of version control of zarr files https://icechunk.io/en/latest/version-control/ Given that it is smoothly integrated with xarray https://icechunk.io/en/latest/xarray/ could be quite cool for the generation of LRA or any drop output

Indeed, you can create a new chunk, commit and then create a new one, and commit again. If any of this operations fails, you can revert the commit. 

Of course this is absolutely hypothetical, some exploration is needed to see if this is feasible

### LINKED ISSUE #2669: [Drop] Option to disable safety checks
### Feature request

There are situations where disabling the safety checks performed by DROP on file integrity would massively speed up the data reduction. This is not meant to be applied in the operational cycle, but could be useful anyhow.

oloapinivad:
## PR description:

This introduces three parallel modules for DROP, one that keeps the usual pipeline in netcdf, and another which support zarr output, and a lastone with icechunk. It also restructure tests and many of CLI options

Close #1888 
Close #2334 

----

 - [x] Changelog is updated.


oloapinivad:
So after a bit of working I have a decent configuration. The current structure introduce a BaseWriter class that can be populated by both a NetCDF engine and by Zarr engine. Tests are increased but coverage is not exceptionally high due to many execptions. 

The main question for long term development is the current month by month approach, which I replicated in zarr to preserve a failsafe approach, is actually effective from the computational aspect. We produce yearly files of zarr for each variable, will this actually work? Perhaps there is a way to do this threadsafe with parallel process, but has to be explored. With the current approach however we can introduce a new writer overriding some of the base features, and this should make the trick. 

Waiting for review

oloapinivad:
I have just added a writer for icechunk, that using the commit approach should be failsafe. However, this is not supported by intake so there is a limitation in the application for AQUA catalogs

oloapinivad:
I think it is working. It would be nice to have a moment to verify that most the solution implemented are efficient.

mnurisso:
> I think it is working. It would be nice to have a moment to verify that most the solution implemented are efficient.

Can you update the template? It is not aligned to the documentation. I'm ready to test it on the data lake.

oloapinivad:
> > I think it is working. It would be nice to have a moment to verify that most the solution implemented are efficient.
> 
> Can you update the template? It is not aligned to the documentation. I'm ready to test it on the data lake.

I did not touch anything in the template nor the docs, there are also too many options, we might want to make a reasoned cleaning

oloapinivad:
Various fixes, compression for zarr and icechunk, CLI can run without configuration file. 
I would stop here, more will come in the future but we can merge this one to checkpoint

silviacaprioli (approved):

aqua/core/console/drop.py
```diff
@@ -83,53 +128,70 @@ def drop_execute(args):
     print("AQUA version is: " + version)
 
     file = get_arg(args, "config", "drop_config.yaml")
```
mnurisso, line 130: Do we really use the `drop_config.yaml` as a default?
What if we drop a default here and if file is None you avoid the try except structure at all?
It is not clear to me the complexity of this structure at the moment

aqua/core/drop/catalog_entry_builder.py
```diff
@@ -167,21 +175,21 @@ def create_entry_details(self, basedir=None, catblock=None, driver="netcdf", sou
...
             self.logger.info("Updated urlpath in existing catalog entry to %s", catblock["args"]["urlpath"])
 
-        if driver == "netcdf":
+        if output_format == "netcdf":
+            #    self.logger.warning("Setting xarray_kwargs for NetCDF driver")
             catblock["args"]["xarray_kwargs"] = {"decode_times": True, "combine": "by_coords"}
 
-            # Jinja parameters to be replaced in the urlpath
-            jinja_params = {"realization": self.realization, "region": self.region, "stat": self.stat}
+        # elif output_format == "zarr":
```
mnurisso, line 183: What is this? A remnant or a future TODO?

aqua/core/drop/drop_writer_base.py
```diff
@@ -0,0 +1,618 @@
...
+
+    Implements common logic for monthly file/store creation and yearly concatenation.
+    Subclasses must implement format-specific methods.
+    """
+
+    def __init__(
+        self,
+        tmpdir,
+        outdir,
+        filename_builder=None,
```
mnurisso: can be useful for maintenance to explicity the type here since it is a custom one

aqua/core/drop/drop_writer_base.py
```diff
@@ -0,0 +1,618 @@
...
+    ):
+        """
+        Initialize base writer.
+
+        Args:
+            tmpdir: Temporary directory for intermediate files
+            outdir: Output directory for final files
+            filename_builder: OutputPathBuilder instance for filename generation
+            loglevel: Logging level
+            _chunk_stats: the list of chunk stats dictionaries, each with keys:
```
mnurisso: these are not in the init, am I missing something?

aqua/core/util/util.py
```diff
@@ -43,11 +43,11 @@ def get_arg(args, arg, default):
     Returns:
         The argument value or the default value
     """
-
-    res = getattr(args, arg)
-    if not res:
-        res = default
-    return res
+    return getattr(args, arg, None) or default
+    # res = getattr(args, arg)
```
mnurisso: remove comment

aqua/core/util/util.py
```diff
@@ -65,14 +65,14 @@ def extract_attrs(data, attr):
...
 
-def username():
-    """
-    Retrieves the current user's username from the 'USER' environment variable.
-    """
-    user = os.getenv("USER")
-    if user is None:
-        raise EnvironmentError("The 'USER' environment variable is not set.")
-    return user
+# def username():
```
mnurisso: can be deleted

docs/sphinx/source/drop.rst
```diff
@@ -49,6 +49,9 @@ DROP's architecture enables various data processing tasks:
 
 DROP can be explored in the `DROP notebook <https://github.com/DestinE-Climate-DT/AQUA/blob/main/notebooks/drop/drop.ipynb>`_.
 
+.. note ::
+    DROP is designed to be flexible and can be used for a wide range of data reduction tasks beyond the specific use cases mentioned above.
```
mnurisso, line 54: I think an empty line is needed for correct formatting

docs/sphinx/source/drop.rst
```diff
@@ -477,6 +477,13 @@ Usage
 
 **Options:** these override the configuration file options.
 
+.. note::
```
mnurisso, line 481: Is ok now, could be made more evident in the future

mnurisso (approved):
No major comment, we can merge it

PR #2829, opened 2026-04-22
Fix NoDataError call

### LINKED ISSUE #2828: NoDataError call with wrong arguments
### machine

Lumi

### catalog

ci

### version

main

### What happened?

In a situation where the CI data have not yet been installed, the following should raise a NoDataError, but it fails because it is called with two arguments instead of one, see line 196 of the reader where we have 

````
            raise NoDataError(
                    f"No NetCDF files available for {self.model} {self.exp} {self.source}, ",
                    f"please check the urlpath: {self.esmcat.urlpath}",
                )
````

and this should be 

````
            raise NoDataError(
                    f"No NetCDF files available for {self.model} {self.exp} {self.source}, " +
                    f"please check the urlpath: {self.esmcat.urlpath}"
                )
````

```python
from aqua import Reader
reader = Reader(model="IFS", exp="test-tco79", source="short")
```

### Error messages

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/pfs/lustrep4/users/cosnerdi/AQUA/aqua/core/reader/reader.py", line 196, in __init__
    raise NoDataError(
          ^^^^^^^^^^^^
TypeError: NoDataError.__init__() takes from 1 to 2 positional arguments but 3 were given

None

CostanzaD:
## PR description:

This fixes a wrong call to the NoDataError exception 

Close #2828 

 - [x] Changelog is updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Unreleased in the current development version (target v1.0.0):
 Main changes:
 
 Complete list:
+- Small fix to wrong arguments of exception call (#2829)
 - Adding LLM directives in `AGENTS.md` (#2820)
 - Add `Ruff` linter and formatter to CI and add `pre-commit` implementation (#2748, #2786, #2791, #2815)
 - Fix race condition during parallel tests (#2805)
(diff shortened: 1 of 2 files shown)


jhardenberg (approved):
great!

PR #2832, opened 2026-04-24
[Hotfix] CatGen: more flexible handling of forcing string 

silviacaprioli:
## PR description:

I updated the logic so that the forcing string is now built automatically, including the prefix (e.g., baseline, projections), without hardcoding specific experiment names. This makes it easier to support new storylines or scenarios without needing to update the forcing map each time.

 - [x] Changelog is updated.

### DIFF
--- a/src/aqua/cli/catgen.py
+++ b/src/aqua/cli/catgen.py
@@ -205,6 +205,15 @@ class AquaFDBGenerator:
         if not result:
             raise ValueError(f"Unexpected {value_type}: {value}")
         return result
+    
+    @staticmethod
+    def get_forcing(experiment: str, prefix_map: dict) -> str:
+        """Get the forcing string based on the experiment name."""
+        normalized = re.sub(r"[^a-z0-9]", "", experiment.lower())
+        for key, prefix in prefix_map.items():
+            if experiment.startswith(key):
+                return f"{prefix}-{normalized}"
+        return normalized
 
 
     def load_jinja_template(self, template_file):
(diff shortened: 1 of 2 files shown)


src/aqua/cli/catgen.py
```diff
@@ -344,17 +353,15 @@ def create_catalog_entry(self, all_content):
             
             resolution_id = self.get_value_from_map(self.config['resolution'], resolution_map, 'resolution')
 
-            forcing_map = {
-                'hist': 'baseline-hist',
-                'cont': 'baseline-cont',
-                'SSP3-7.0': 'projections-ssp370',
-                'Tplus2.0K': 'tplus2K'
+            prefix_map = {
```
oloapinivad: there might be abrupt4CO2 experimens, will they work as well?

src/aqua/cli/catgen.py
```diff
@@ -344,17 +353,15 @@ def create_catalog_entry(self, all_content):
...
+                "hist": "baseline",
+                "cont": "baseline",
+                "SSP": "projections",
```
oloapinivad: storylines are threated differently?

oloapinivad:
On extra thing, I know it is boring but perhaps testing this with all the possible cases given that it is going to important in the workflow could be a good option

src/aqua/cli/catgen.py
silviacaprioli: If there will be a specific prefix for it we will need to add it, otherwise it will result in
  `abrupt4CO2`->`abrupt4co2`

src/aqua/cli/catgen.py
silviacaprioli: you are right, "story-nudging" should be added

src/aqua/cli/catgen.py
silviacaprioli: Ah, I see there is already one of them [baseline-abrupt4xco2.yaml](https://github.com/DestinE-Climate-DT/Climate-DT-catalog/blob/main/catalogs/climatedt-gen2/catalog/ICON-5km/baseline-abrupt4xco2.yaml)
They should have "baseline" as prefix

src/aqua/cli/catgen.py
oloapinivad: cant we get this from activity key? i guess this is not a parameter that user or workflow read, correct? 

the risk is that any other configuration will not be covered, @mnurisso what do you think?

silviacaprioli:
Actually, I’m realizing that the filename (baseline-cont, baseline-hist, etc.) is determined solely by `config['exp']`.
So the user can choose whatever name they want.
All that mapping was just to write the forcing into main.yaml.
So I'm no longer sure how we want to proceed...

silviacaprioli:
Actually, at the last aqua meeting we decided to keep the possibility for the user to be completely free to choose whatever name they want in `config['exp']`. 
So it will be up to the user to follow the correct convention for operational runs.

In this PR, I will therefore limit myself to generale the way the forcing is extracted, it’s no longer as critical as it initially seemed.

silviacaprioli:
Codecov is failing, but I would go ahead and merge this PR anyway. We can open another PR to improve the tests for catgen.py, since coverage is currently only at 30%.

mnurisso (approved):

PR #2834, opened 2026-04-27
[DP] `aqua analysis` templating and `--kind` option

### LINKED ISSUE #2814: Correct the `--threads` option in aqua analysis
### Feature request

This is actually a misnomer since it controls the number of parallel diagnostics, it has to be fixed

### LINKED ISSUE #2836: AQUA analysis cannot run if diagnostics are not installed
### Feature request

Currently, aqua analysis call are actually expecting to find aqua diagnostics. In the view of our modular approach, this will not work. In #2834 I am adding an error trap to verify that at least tests works, but this is probably not enough.

More in general, we should probably provide a template for a configuration file of aqua-analysis and a mock cli to be run in aqua core if we want this feature to be really exploited. 

Alternatively, we should move aqua analysis to the diagnostics

oloapinivad:
## PR description:

This PR introduces the possibility of doing templating via temporary configuration files generated on the fly by aqua analysis. A `experiment_kind_file` is read by the tool and then used to configure different j2 files.

This is a companion PR of https://github.com/DestinE-Climate-DT/AQUA-diagnostics/pull/230 

 - [x] Changelog is updated.


aqua/core/analysis/analysis.py
```diff
@@ -142,17 +153,20 @@ def run_diagnostic_func(
...
         # Build conditional arguments
         if regrid:
             extra_args += f" --regrid {regrid}"
 
-        if parallel:
-            nworkers = tool_config.get("nworkers")
-            if nworkers is not None:
-                extra_args += f" --nworkers {nworkers}"
+        if not serial:
+            tool_nworkers = tool_config.get("nworkers")
```
mnurisso, line 164: I'd like a comment on what they are to avoid confusion in the future

aqua/core/analysis/analysis.py
```diff
@@ -170,17 +184,76 @@ def run_diagnostic_func(
...
+                collection=collection,
+                tool=tool,
+                script_path=cli_path,
+                extra_args=args,
+                loglevel=loglevel,
+                logger=logger,
+                logfile=logfile,
             )
 
+        if exp_kind_dict:
```
mnurisso, line 209: comment about what is done here. If I understood this is to remove a temporary file created by the jinja replacement? I'd like this comment here if I understood correctly

aqua/core/analysis/analysis.py
```diff
@@ -170,17 +184,76 @@ def run_diagnostic_func(
...
+            temp_cfg_dir = os.path.dirname(cfgs[0])
+            shutil.rmtree(temp_cfg_dir, ignore_errors=True)
+            logger.debug("Removed temporary config directory: %s", temp_cfg_dir)
+
+
+def configure_experiment_kind(exp_kind, exp_kind_file, logger):
```
mnurisso, line 215: Long term, but since you're passing the logger to every function I have the feeling a class would be ideal here

aqua/core/analysis/analysis.py
```diff
@@ -170,17 +184,76 @@ def run_diagnostic_func(
...
+    """ "
```
mnurisso: extra " left

aqua/core/analysis/analysis.py
```diff
@@ -170,17 +184,76 @@ def run_diagnostic_func(
...
+    Configure the experiment kind based on the provided kind and configuration file.
```
mnurisso, line 217: Docstring missing

aqua/core/analysis/analysis.py
```diff
@@ -170,17 +184,76 @@ def run_diagnostic_func(
...
+        return None
+
+    if not os.path.exists(exp_kind_file):
+        raise FileNotFoundError(f"Experiment kind config file '{exp_kind_file}' not found.")
+
+    logger.info(f"Configuring experiment kind: {exp_kind} using config file: {exp_kind_file}")
+    complete_dictionary = load_yaml(exp_kind_file)
+
+    if exp_kind not in complete_dictionary:
+        logger.warning(f"Experiment kind '{exp_kind}' not found in config file '{exp_kind_file}'. Default selected")
```
mnurisso: logger.error maybe?

docs/sphinx/source/aqua-analysis.rst
```diff
@@ -76,31 +76,45 @@ so that the script can be used in a batch job or in a workflow. These override c
...
     The output directory to use.
     The default in the config file is ``$AQUA/cli/aqua-analysis/output``.
     Better to use an absolute path.
 
-.. option:: -l <loglevel>, --loglevel <loglevel>
+.. option:: -k <kind>, --kind <kind>
 
-    The log level to use for the cli and the diagnostics.
-    Default is ``WARNING``.
+    Experiment kind to run (e.g. ``historical``, ``scenario``, ``storyline``).
```
mnurisso, line 87: It would be cool to have here, or anywhere else the full list of kind, which I imagine is now these three.

mnurisso (approved):

aqua/core/analysis/analysis.py
oloapinivad: I am quite sure those are never used btw

aqua/core/analysis/analysis.py
oloapinivad: 100% agree

docs/sphinx/source/aqua-analysis.rst
oloapinivad: Well we do not know this since we are in aqua-core :-)

mnurisso: Ok, but I guess no documentation update was don in aqua-diagnostics, right?

aqua/core/analysis/analysis.py
mnurisso: so let's remove them 😶‍🌫️

mcadau:
https://jinja.palletsprojects.com/en/stable/api/#jinja2.StrictUndefined

reference on what we were discussing about StrictUndefined

oloapinivad:
Following discussion with @mcadau we added an option for automatic raise `UndefinedError` if a jinja tag is undefined instead reaplcing with a blank line. Only problem that the change is applied to `load_yaml` so that it might have consequences also to other part of the code, let's hope test will identify it.

oloapinivad:
Tests are failing since in some occasion the `load_yaml` call is done with jinja files with incomplete definitions. we probably need to have a flag to control the undefined option, even if this will implies reinitializing everytime the loader. 
perhaps we could define two loaders?

oloapinivad:
Introduced a mixed jinja env call, test are passing, we might need a better test but we are there

aqua/core/analysis/analysis.py
oloapinivad: I would target this in #2857

aqua/core/analysis/analysis.py
oloapinivad: To be done in #2857

aqua/core/analysis/analysis.py
oloapinivad: done

aqua/core/analysis/analysis.py
oloapinivad: done

docs/sphinx/source/aqua-analysis.rst
oloapinivad: in the aqua-core documentation we have 

```rst
.. option:: -k <kind>, --kind <kind>

    Experiment kind to run (e.g. ``historical``, ``scenario``, ``storyline``).
    This works in combination with the jinja template configuration files available in AQUA-diagnostics, and allow for configuring the diagnostics startdate/enddate and other parameters based on the experiment kind.
    Overrides the value from the configuration file.
```

not sure if this enough, but it is not completely forgotten.

mcadau (approved):

docs/sphinx/source/aqua-analysis.rst
oloapinivad: resolving this, please open an issue in case of need

PR #2843, opened 2026-05-05
Intake migration to v2

jhardenberg:
## PR description:

First attempt at making our current catalogs and reader work with intake and intake-xarray >= 2.0.0

----

 - [x] Changelog is updated.
 - [x] Run Cross check tests for AQUA-diagnostics [AQUA-diagnostics cross-check workflow](https://github.com/DestinE-Climate-DT/AQUA-diagnostics/actions/workflows/aqua-diagnostics-cross-check.yml)

### DIFF
--- a/aqua/core/reader/reader.py
+++ b/aqua/core/reader/reader.py
@@ -192,6 +192,11 @@ class Reader:
 
         # Manual safety check for netcdf sources (see #943), we output a more meaningful error message
         if isinstance(self.esmcat, intake_xarray.netcdf.NetCDFSource):
+            # HACK to get expanded urlpath and metadata for netcdf sources
+            _, oa = self.esmcat._entry._create_open_args(self.kwargs)
+            self.esmcat.urlpath = oa["urlpath"]
+            self.esmcat.metadata = oa["metadata"]
+
             if not files_exist(self.esmcat.urlpath):
                 raise NoDataError(
                     f"No NetCDF files available for {self.model} {self.exp} {self.source}, "
@@ -1087,7 +1092,7 @@ class Reader:
         """
 
         # list available files in folder
-        files = sorted([f for x in esmcat.urlpath for f in glob(x)])
+        files = sorted([f for x in esmcat._entry._open_args["urlpath"] for f in glob(x)])
         self.logger.debug("Total files before filtering: %s", len(files))
 
         # this will consider only files that have "year" in their filename
@@ -1103,12 +1108,17 @@ class Reader:
             raise ValueError(f"Filter type {filter_key} not recognized.")
 
         # replace the urlpath with the filtered one searching the regex
-        esmcat.urlpath = [f for f in files if any(p.search(os.path.basename(f)) for p in pattern)]
+        esmcat._entry._open_args["urlpath"] = [f for f in files if any(p.search(os.path.basename(f)) for p in pattern)]
 
-        if len(esmcat.urlpath) == 0:
+        if len(esmcat._entry._open_args["urlpath"]) == 0:
             raise NoDataError("No files found after filtering the catalog!")
 
-        self.logger.debug("Selected: %s files from %s to %s", len(esmcat.urlpath), esmcat.urlpath[0], esmcat.urlpath[-1])
+        self.logger.debug(
+            "Selected: %s files from %s to %s",
+            len(esmcat._entry._open_args["urlpath"]),
+            esmcat._entry._open_args["urlpath"][0],
+            esmcat._entry._open_args["urlpath"][-1],
+        )
 
         return esmcat
 
--- a/aqua/core/reader/reader.py
+++ b/aqua/core/reader/reader.py
@@ -723,7 +723,8 @@ class Reader:
     def intake_user_parameters(self):
         """Lazy loader for intake user parameters to avoid expensive describe() calls."""
         if not hasattr(self, "_intake_user_parameters"):
-            self._intake_user_parameters = self.esmcat.describe().get("user_parameters", {})
+            # self._intake_user_parameters = self.esmcat.describe().get("user_parameters", {})
+            self._intake_user_parameters = [v.describe() for v in self.esmcat.cat.user_parameters.values()]  # intake2 change
         return self._intake_user_parameters
 
     def _filter_kwargs(self, kwargs: dict = {}, engine: str = "fdb", intake_vars: dict = {}, databridge: str = None) -> dict:
@@ -927,7 +928,8 @@ class Reader:
         # attribute. I would not add it since it is a deprecated feature
         if fdb_var is None:
             self.logger.warning("No 'variables' metadata defined in the catalog, this is deprecated!")
-            fdb_var = esmcat.describe()["args"]["request"]["param"]
+            # fdb_var = esmcat.describe()["args"]["request"]["param"]
+            fdb_var = esmcat._entry._open_args["request"]["param"]  # This does work with intake2
             fdb_var = to_list(fdb_var)
 
         # We avoid the following loop if the user didn't specify any variable
--- a/aqua/core/reader/reader.py
+++ b/aqua/core/reader/reader.py
@@ -1176,12 +1176,14 @@ class Reader:
 
             esmcat.xarray_kwargs.update({"decode_times": coder})
 
-        read_kwargs = getattr(esmcat, "xarray_kwargs", {}).copy()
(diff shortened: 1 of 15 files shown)


jhardenberg:
So, the very interesting result is that all tests related to the GSV driver work flawlessly. All failures are only related to reading netcdf!

oloapinivad:
I already saw this errors in my tests, it should be enough to add

```yaml
xarray_kwargs:
        engine: netcdf4
```

to each of the source you load.

jhardenberg:
Let me see if it is possible to do something to override this at reader level

jhardenberg:
Note: I had to explicitly `import xarray_intake.xzarr` becouse the submodule is commented out among intake_xarray's exports

jhardenberg:
We are down to just three failures/errors!

jhardenberg:
The test was failing because gribscan is not a valid module for intake anymore. I created this issue as a reminder https://github.com/DestinE-Climate-DT/Climate-DT-catalog/issues/248
This is a problem only for the levante catalog ....

For our purpouses here I just changed the catalog used in the console test

jhardenberg:
All tests pass! We have a working migration using the original catalogs.
The only thing left to do is to verify intake-esm but I would do that in a separate issue

aqua/core/reader/reader.py
```diff
@@ -188,14 +189,33 @@ def __init__(
...
         self.logger.debug("Using filtered kwargs: %s", self.kwargs)
-        self.esmcat = self.expcat[self.source](**self.kwargs)
+        self.esmcat = self.expcat._entries[self.source](
+            **self.kwargs
+        )  # HACK for intake2 following https://github.com/intake/intake-xarray/issues/150
+
+        if isinstance(self.esmcat, intake_xarray.netcdf.NetCDFSource) or isinstance(
+            self.esmcat, intake_xarray.xzarr.ZarrSource
+        ):
+            # HACK convenience to get expanded url, xarray_kwargs and metadata for netcdf/zarr sources for intake2
```
oloapinivad, line 199: I am wondering if we could generalize this, if not `intake_gsv` source then do this. I mean, a parquet source should behave in the same way as the one here, no?

aqua/core/reader/reader.py
```diff
@@ -188,14 +189,33 @@ def __init__(
...
+            # this provides direct access to the intake netcdf data object
+            self.esmcat.data = self.esmcat.reader.kwargs["args"][0]
+
+            self.esmcat.metadata = self.esmcat.reader.metadata
+            self.esmcat.xarray_kwargs = self.esmcat._entry._captured_init_kwargs.get("args", {}).get("xarray_kwargs", {})
 
-        # Manual safety check for netcdf sources (see #943), we output a more meaningful error message
         if isinstance(self.esmcat, intake_xarray.netcdf.NetCDFSource):
-            if not files_exist(self.esmcat.urlpath):
+            # HACK: Manually expand globs to ensure xarray/intake2 always receives an explicit list of files.
```
oloapinivad, line 208: this means that intake2 cannot expand wildcard?

aqua/core/reader/reader.py
```diff
@@ -1084,29 +1106,36 @@ def _filter_netcdf_files(self, esmcat, filter_key="year"):
...
         # this will consider only files that have "year" in their filename
         # within the startdate and enddate range
         if filter_key == "year":
-            if not (self.startdate and self.enddate):
-                return esmcat
-            keys = list(range(pd.Timestamp(self.startdate).year, pd.Timestamp(self.enddate).year + 1))
-            # create regex pattern for each year: only yyyy will be detected
-            pattern = [re.compile(rf"(?<!\d){yr}(?!\d)") for yr in keys]
-
+            if self.startdate and self.enddate:
```
oloapinivad, line 1115: I am wondering if this manual filtering can work directly with intake parameters, do not know why we never tried.

oloapinivad (approved):
That's great! there is a massive amount of hack but it works so who cares :-)

I left a couple of comments, what is unclear to me is only how you achieved the `user_parameters` jinja rendering, can you clarify this for the records? 

Overall, my take on intake2 is that we need to get rid of intake_xarray and develop a intake reader that is used for aqua. We can create methods that wrap up the hacks currently done that we need to use.

I feel we can revisit this in our dedicated meeting, we can start drafting the ideas maybe before but I refactor the reader capabilities in a modular way.

aqua/core/reader/reader.py
jhardenberg: Actually we already had this expansion in  `_filter_netcdf_files()`, it is just moved up here so that it applies always (not only if a filter key is specified). 
The reason for this  is actually `Xarray.open_mfdataset()` itself, not intake2: if we provide as url a LIST of glob patterns (like `["files_a_*.nc", "files_b_*.nc"]`), this is not recognized. `open_mfdataset()` can only open glob patterns if a single string is provided. In addition there is a small qwirk of intake2:  Indeed if a single url is provided as a string and it contains a glob pattern this is correctly passed to `open_mfdataset`, but if a list with only one element is passed (so a list with one elemnt with an url with a glob pattern) then `open_dataset()` is called and it will not recognize the pattern. So between the Xarray issue and the intake2 one it is better to simply expand directly always.

aqua/core/reader/reader.py
jhardenberg: We could try....  for metadata and access to the data (for the url) it accesses instance variables of the reader itself, so it should work regardless of the driver. It also provides `esmcat.xarray_kwargs` which instead is specifc for xarray, but as it is writte it would not harm to try also with another driver (it woul creat an empty dict).

I am still not sure what happens with intake-esm (to be tested separately from this PR)....

aqua/core/reader/reader.py
jhardenberg: Uh, you will need to explain this directly to me.

jhardenberg:
>  what is unclear to me is only how you achieved the user_parameters jinja rendering, can you clarify this for the records?

First, I found that the ony way to see the rendered url was accessing the data object and its `url` variable. Other methods to apparently see the url (like `._entry` ) show the ORIGINAL url, still with the template.

Second, it is quite crucial for user parameters to be correctly used to use the hack suggested by Martin in https://github.com/intake/intake-xarray/issues/150 , that is why we have the line

`self.esmcat = self.expcat._entries[self.source]( **self.kwargs )`

aqua/core/reader/reader.py
jhardenberg: On the other hand, currently the convenience variables `esmcat.dataset, esmcat.metadata` and `esmcat.xarray_kwargs` are used in the reader code  only if the reader is intake-xarray, so these if statements are justified (even if it would not harm to define them). We could change this in the future if we support (and test) other drivers ....

aqua/core/reader/reader.py
oloapinivad: Given that the intake jinja expansion is working partially, I am wondering if we could override that so that we could provide a list of kwargs that is expanded in the urlpath instead of doing this tricky business here. 

My feeling - in general - is that given the limit that intake has and will have, we could develop it to accomodate it our needs. This is why I would get rid of intake-xarray

aqua/core/reader/reader.py
oloapinivad: thanks for the explanation, this is one of the feature we need to generalize so that it can operates not only on netcdf but on any source with multiple files

aqua/core/reader/reader.py
oloapinivad: this looks another case to get rid of intake-xarray :-)

pyproject.toml
```diff
@@ -40,16 +40,16 @@ dependencies = [
...
     "cartopy>=0.24.1",
     "cfgrib", #enable support for GRIB files
     "dask[distributed]",
     "fastparquet", #enable support for parquet files
     "filelock",
     "gcsfs", # Google Cloud Storage File System: required by ARCO
-    "astropy-healpix",
-    "intake==0.7.0", # Version 2 is a major refactor still in development
-    "intake-xarray<=0.7.0", # Version 2 is a major refactor still in development
+    "intake", # Version 2 is a major refactor still in development
```
oloapinivad: I think we need to set a minimum intake also here, no?

jhardenberg: good catch

oloapinivad:
Test passed, merging this!

PR #2847, opened 2026-05-06
Bump eccodes to 2.47.0

silviacaprioli:
## PR description:
Bump eccodes from 2.41.0 to 2.47.0

Please leave the checkboxes that apply to this pull request.
If you find a missing checkbox, please add it to the list.
Make sure to complete the checkboxes before applying the "ready to merge" label.
Please apply the "run tests" label if you want to trigger CI tests.
 
 - [x] Changelog is updated.

PR #2853, opened 2026-05-10
Extend aqua-core compatibility to python 3.13 and 3.14

oloapinivad:
## PR description:

Starting from #2843 I made a short test when I set up a python matrix for 3.13 and 3.14.
if/when working, the tests have to migrated to more recent python and rebased once #2843 is merged.

@jhardenberg 

 - [x] Changelog is updated

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -12,7 +12,7 @@ dependencies:
   # binaries
   - python>=3.10,<3.13
   - cdo>=2.5.0 #for smmregrid compatibility
-  - eccodes==2.47.0
+  - eccodes==2.41.0
   - cartopy>=0.24.1,<=0.25.0
   - cartopy_offlinedata # this is needed for MN5 HPC
   - netcdf4<=1.7.2
--- a/environment.yml
+++ b/environment.yml
@@ -27,8 +27,8 @@ dependencies:
   - filelock<=3.25.0
   - gcsfs<=2026.2.0
   - healpy<=1.18.1
-  - intake==0.7.0
-  - intake-xarray<=0.7.0
+  - intake
+  - intake-xarray
   - jinja2<=3.1.6
   - kerchunk<=0.2.9
   - matplotlib<=3.10.8
--- a/environment.yml
+++ b/environment.yml
@@ -12,7 +12,7 @@ dependencies:
   # binaries
   - python>=3.10,<3.13
   - cdo>=2.5.0 #for smmregrid compatibility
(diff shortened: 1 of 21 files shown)


oloapinivad:
Ah, it works, super cool!

jhardenberg:
In #2843 I made a small resort of entries in pyproject.toml (astropy-healpix had to be moved to the beginning), so you may need to merge test-intake2 into this one.

jhardenberg:
cool that it works

oloapinivad:
Yes of course, and also #2847 should be taken into account. BTW, have you noticed that test execution time is massively faster with more recent python?

oloapinivad:
I moved everything to python 3.14, and left the aqua-complete monthly check running with all the python version from 3.10 up to 3.14. If working this will be a major update. 

Unfortunately there is the usual mess with the commits, which implies that there is more commit than actually made, but with the squash at the merge they will disappear

oloapinivad:
Test passing, waiting for review

environment.yml
```diff
@@ -10,7 +10,7 @@ channels:
 # please note that some of these dependencies are made explicit on purpose, but are not directly required by conda/mamba
 dependencies:
   # binaries
-  - python>=3.10,<3.13
+  - python>=3.10,<3.15
```
jhardenberg, line 13: Should this not be `<=3.14` ?

pyproject.toml
```diff
@@ -20,7 +20,7 @@ description = "AQUA; a model evaluation framework for high-resolution climate mo
 readme = "README.md"
 keywords = ["climate", "modeling", "diagnostics", "high-resolution"]
 license = {file = "LICENSE"}
-requires-python = ">=3.10"
+requires-python = ">=3.10,<3.15"
```
jhardenberg, line 23: <=3.14?

jhardenberg (approved):
I realize that it was <3.13 already before, but why not <=3.14 for the python pins?

environment.yml
oloapinivad: my confused understanding of pinning is that it is safer to exclude rather than include. There is the risk that setting `<=3.14` implies only `3.14.0` and not `3.14.1`. However, at worst it should not change anything, but we can double check if my understanding is correct

oloapinivad: apparently no distinction :-) https://gemini.google.com/share/2284bf54960d

mnurisso (approved):

PR #2865, opened 2026-05-13
Aqua grid deploy

### LINKED ISSUE #2808: `aqua grids deploy <source_grid_name>` new feature
### Feature request

For the Generation 2 data access we've wrote a simple script that taking the `source_grid_name` deploys the needed grids file. The natural evolution would be to create a console entry point based on that script and make it available for every grid, making it part of the available tools of the console.

[Here](https://github.com/DestinE-Climate-DT/climatedt-community-resources/blob/update-aqua-access/example_aqua/grids_deploy.py) the original script

mnurisso:
## PR description:

Initial implementation.
I've a bit of doubts on how to try to implement the wildcard option, how can I have a wildcard taken as command line argument? It didn't work in my tests, any suggestion is welcome.

Close #2808 

@oloapinivad can you comment on the wildcard?

 - [x] Changelog is updated.


oloapinivad:
I am wondering if you can move the set of core function into a dedicated module to be imported, so that we can later call the grid deployer also from the reader

mnurisso:
> I am wondering if you can move the set of core function into a dedicated module to be imported, so that we can later call the grid deployer also from the reader

Done, if you like the new file position I can start documentation and tests

mnurisso:
Hi all, this is ready for the review. As requested the entry point just uses a call to a deployer class which can be later on added inside the Reader to trigger an automatic download of the necessary grid when the user opens a new source with a not downloaded grid.

Additionally I removed the cli instrument for DKRZ grids download, as it would be a third way to donwload grids and it is only adding noise. I can remove the grids from there if we think it is important to save space there.

Let me know if further modifications are needed.

mnurisso:
Again a polytope running out of time, other tests are fine

silviacaprioli:
Looks good to me, maybe it's worth adding an explicit example in the documentation?

silviacaprioli (approved):

silviacaprioli:
tests failing for connection timed out

mnurisso:
DVC is down, wilma is down. No test is possible

oloapinivad:
@silviacaprioli @mnurisso can we merge this now correct?

silviacaprioli:
I would say yes

mnurisso:
Ok, merging

PR #2869, opened 2026-05-14
add source and target grid as AQUA attributes

### LINKED ISSUE #2722: Missing info for destination grid as attribute
### Discussion

As discussed with @jhardenberg, Histogram (and maybe other diagnostics in the future) would benefit to have the destination regrid informations in the plot description after regridding. To this aim, an attribute in the netCDF would be beneficial after regridding. 
This is not present at the actual state on any netCDF file, and even if we add the possibility to add this attribute on the new netCDFs, the current LRA data don't have any `r100` attribute indicating the destination grid.

At the moment, Jost and I thought we could simply merge [this PR](https://github.com/DestinE-Climate-DT/AQUA-diagnostics/pull/85) without this information available on the description. 

What do you think, in any case, on the possibility to add this attribute?

oloapinivad:
## PR description:

As discussed, I extended the marking of attributes from the regridding operation. I added source and target grid, but this is not very clean for a series of reasons

- In principle, it should be the `Regridder` class to take care of this, not the `Reader`. However, the regridder is built so that `remap_method` and `tgt_grd_name` are argument of the weights generation, and then ignored by the regrid operation. We should access the smmregrid object to get this info
- The `remap_method`,  another useful info, is set to None because is controlled by the `Regridder`class
- The `set_attrs` function is only a function of the reader, and in general it calls for a restructuring of the util set.

Long story short, this is fine, but a more elegant solution will require a bit of reworking and thinking

Close #2722

## People involved:

@mcadau @jhardenberg 

 - [x] Changelog is updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ ClimateDT workflow modifications:
 Main changes:
 
 Complete list:
+- Add AQUA attributes for source and target grid (#2869)
 - Fallback test download from wilma (#2867)
 - push_analysis png metadata support and remove imagemagick dependency (#2866)
 - Support for python 3.13 and 3.14, with new default from 3.12 to 3.14 (#2853)
--- a/aqua/core/reader/reader.py
+++ b/aqua/core/reader/reader.py
@@ -618,7 +618,9 @@ class Reader:
         out = self.regridder.regrid(data)
 
         # set regridded attribute to 1 for all vars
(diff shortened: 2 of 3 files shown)


oloapinivad:
All set, ready to review

silviacaprioli (approved):

silviacaprioli:
looks good to me, @mcadau do you have further comments? otherwise I would merge it

mcadau:
I agree, looks good to me. Merging

PR #2873, opened 2026-05-15
Update paths for load container script

### LINKED ISSUE #2532: Load container script to be updated
The container load should be available in both the core and diagnostic repo, pointing to the respective folders and file names.

To be discussed the backward compatibility of the script for versions previous than v0.21.0

tovaz92:
The paths with the locations for the container loader needs to be updated.

Close #2532 

 - [ ] Changelog is updated.


jhardenberg:
I am not sure that this is the right solution. Sometimes we will use only the core (e.g. drop does it). so it makes sense for this script to be distributed with the core. It woud rather be useful if it were more generic and capable of loading both the aqua-core and the aqua-diagnostics container, depending if the fuctionality of the latter is needed

oloapinivad:
> I am not sure that this is the right solution. Sometimes we will use only the core (e.g. drop does it). so it makes sense for this script to be distributed with the core. It woud rather be useful if it were more generic and capable of loading both the aqua-core and the aqua-diagnostics container, depending if the fuctionality of the latter is needed

I don't understand this comment sorry. We agreed yesterday that we will remove the load-container from aqua-core and keep it in aqua-diagnostics, why this second thought?

mnurisso:
Nobody will deploy the container in any machine mentioned in the script. I don't see the point of changing strategy here. In the climatedt machines the diagnostic container will be the one devoted to all aqua tasks in my opinion

tovaz92:
So to recap, I think we should briefly make a decision over this.
Linking also to our internal discussion, one approach is to move both the `aqua-web` and `drop` folders into AQUA-diagnostics (making the necessary adaptations to make it work).
On the other hand, another solution discussed yesterday was to move the`load-aqua-container.sh` from AQUA to AQUA-diagnostics. I propose to discuss this briefly here to move forward once decided.

CostanzaD:
Is there any decision about this one?

tovaz92:
At the moment, the loader-aqua-container script is used in core (by aqua_drop.j2, aqua-web.j2; but these need aqua analysis to run, although load-aqua-container is meant to run with the diagnostics). 
So, how to approach this? 
- Shall we make 2 copies? 
- Or do we just put the loader in aqua diagnostics? 
- we keep everything cli-related into the core, and we allow the possibility to leave the script in the core but with a flag to call the load container of aqua-diagnostics (this last would follow @jhardenberg suggestion, in this way it would be needed to clone the core repo to access the script, but it is a fair compromise)? 
-> Note, nobody will use the container of the core to run aqua analysis, as will use the container of aqua-diagnostics.

In `aqua/core/cli` some stuff is legacy, some not:
- aqua-container : ok, keep
- drop: script extra
- hpc2020-install: ok     
- orca-grids: semi-deprecated
- aqua-web: Should this be in `aqua/core` instead? We are using this a lot (as well as the workflow)
- eccodes: deprecated
- lumi-install: ok        
- orography_from_data: ok
- benchmarker: deprecated
- grid-from-data: almost deprecated since now it is mainly inside the Gridbuilder
- nc2zarr: ok

oloapinivad:
My suggestion is going along the lines of the third bullet point option. the 'cli' folder is a patchwork of different elements that we use or not (I would also clean this up), we could decide that all the CLI stays in `aqua-core` and nothing but the installation procedures (I have some doubt honestly about those too) can go to `aqua-diagnostics`

on top of this, @jhardenberg can we think about moving aqua-web as aqua/core component?

mnurisso:
This code is probably going to be better exploited in a dedicated aqua-climatedt repo

jhardenberg:
Since I needed it, I made an alternative version (expanded with some additional functionality and which can install both core and diagnostics containers) in #3043
Of course that could be moved to diagnostics if we wish

jhardenberg:
I am closing this one because it is supersed by https://github.com/DestinE-Climate-DT/AQUA/pull/3043

PR #2897, opened 2026-06-09
AQUA analysis as a class

### LINKED ISSUE #2857: Revisit dask cluster option in aqua-analysis and in other elements (e.g. DROP)
### Feature request

Currently aqua-analysis is configuring multiple dask option, from the `local_cluster`, to the number of workers and threads, but we are not fully sure that all of them are passed to the cluster and then to the diagnostic. For example, there is no control that the total number of workers used by the tool combined exceeds the available workers in the current cluster.

Function as `run_diagnostic_collection` receive many arguments, and might benefit of a class to correctly pass all the arguments and control.

On top of that, we noticed that the function that open an existing cluster is currently into the diagnostic module, obliging a new AQUA-friends developer to connect himself to the existing cluster. Especially in the context of the development of AQUA-friends, this will be an useful refactoring.

oloapinivad:
## PR description:

Initial effort to transform aqua analysis into a class rather than a collection of methods: will later take care of putting an order in methods and different options. 

Close #2857 

----

 - [x] Changelog is updated.


oloapinivad:
I worked toward a simplifcation of the class, using a few helper method to set the required keys so that now `model`, `exp` and etc. keys does not need to passed to each call of the `run_diagnostic_collection` but are part of it.
I also make use of the `run_command`for both checker and tool. Probably a better solution is to have a specific class to handle each collection, but this seems to me quite a good initial upgrade. 

I suspect this might cause some small disruption in the complete pipeline and we will need to test it in practical usage on a full experiments since tests are only mocking usage.

oloapinivad:
I somehow the refactor of aqua analysis. 
- `Analysis` class now handles all the different steps, store them in internal attributes. A group of `set_*` method is set to configure the different options and to guard against possible failure. 
- A `DaskCluster` class has been introduce as a separate module. This is used by `Analysis` but it will be refined in the future to be used by `Drop` and possibly also by diagnostics. Of course, some adjustment are likely needed. The class is used within `Analysis`. 
- The CLI is now using the `Analysis` class extensively
- A template for `config.aqua-analysis.tmpl` has been introduced
- Tests has been refactored accordingly, and some further check for the cli checker has been added
- Command args have bene cleaned and documentation modified accordingly. 

Ready for review. Before merging I would like to run it on real data, but before we should fix issues we are having with the main and with the jinja templating.

oloapinivad:
this has been forgotten, what do we do? is ready since more than one month...

mcadau:
Well I can review it if you want. The two questions are: 
1) in this period something has dramatically changed in the main enough to think merging this can be disruptive in some way?
2) how the (maybe close) merge of aqua-backend deals with aqua analysis becoming a class? Nothing, I guess, so that the two are "totally" independent with each other? Are we sure about this?

oloapinivad:
> 1. in this period something has dramatically changed in the main enough to think merging this can be disruptive in some way?

don't think is the case

> 2. how the (maybe close) merge of aqua-backend deals with aqua analysis becoming a class? Nothing, I guess, so that the two are "totally" independent with each other? Are we sure about this?

checking the modified files, it does not seem that we should foresee relevant conflicts

silviacaprioli:
Yesterday we said that we would review it and merge it after the release, no?

tovaz92:
FYI the v1.0.0a6 has been released.

oloapinivad:
Can we merge it before v1.0.0?

aqua/core/dask/daskcluster.py
```diff
@@ -0,0 +1,94 @@
...
+        return self._address
+
+    @property
+    def active(self):
+        """Return True if the cluster is active."""
+        return self._cluster is not None
+
+    def setup(self, nworkers: int, nthreads: int, mem_limit: str, connect_timeout: float = None, tcp_timeout: float = None):
+        """
+        Spin up a LocalCluster with the specified configuration if not already running.
```
mnurisso: docstring more expanded to be added

aqua/core/dask/daskcluster.py
```diff
@@ -0,0 +1,94 @@
+"""Centralised manager for dask cluster in AQUA"""
```
mnurisso, line 1: isn't missing the init file for this module?

aqua/core/dask/daskcluster.py
```diff
@@ -0,0 +1,94 @@
...
+        if not self.active:
+            self.logger.debug("No active cluster to close.")
+            return
+
+        self.logger.info("Closing dask cluster at %s.", self._address)
+        self._cluster.close()
+        self._cluster = None
+        self._address = None
+
+    def _configure_timeouts(self, connect_timeout: float = None, tcp_timeout: float = None):
```
mnurisso, line 86: docstrings are wrong

aqua/core/dask/daskcluster.py
mnurisso, line 1: Since the cluster open/close code for the diagnostic seems quite similar to this one can we be sure that this code is easily used by diagnostics? This would simplify and move to the correct repo that chunk of code.

aqua/core/analysis/cli_checker.py
mnurisso, line 1: With @tovaz92 we were wondering if this should become a project dependent tool, imagine a project accessing only with a specific driver with specific kwargs or not regridding or other options. The checker should become part of a dictionary like the tools specifying where to find it depending on the project using the analysis.

This of course to be done in a follow up PR.

silviacaprioli:
> Can we merge it before v1.0.0?

Has it been tested with a full run of aqua-analysis?

mnurisso (approved):

aqua/core/dask/daskcluster.py
oloapinivad: we cannot yet, but this is the underlying idea to have a centralised tool for anybody working with dask.

aqua/core/analysis/cli_checker.py
oloapinivad: This is discussed here https://github.com/DestinE-Climate-DT/AQUA/issues/2973, correct?

aqua/core/dask/daskcluster.py
oloapinivad: fixed

aqua/core/dask/daskcluster.py
oloapinivad: fixed

aqua/core/dask/daskcluster.py
oloapinivad: fixed

silviacaprioli (approved):

silviacaprioli:
tested successfully with aqua analysis. Can we merge it?

mcadau:
A(nother) full AQUA analysis run is passing and with the same run time as `main`, so we can finally merge this

PR #2899, opened 2026-06-09
DROP regrid_first flag introduction

### LINKED ISSUE #2890: DROP `regrid_first` flag
### Feature request

We want to implement a `regrid_first` flag for DROP. This is necessary because by default the regrid + area selection is performed after the other operations.

This is mostly fine, but it does not work when the Timstat operation disrupt the longitude and latitude coordinates, e.g. for the histograms.

The flag will be exposed at every level for DROP.

Suggestion is to work taking the evaluation method currently on DROP and splitting it on other smaller methods, so that possible further modifications of the operations order may be applied in a simple way.

mnurisso:
## PR description:

Introducing a `regrid_first` flag for time statistics that disrupt the lon-lat coordinates.
While doing this the different operations are split in hidden methods.

Close #2890 

 - [x] Changelog is updated.


aqua/core/drop/drop.py
```diff
@@ -554,6 +562,39 @@ def _remove_regridded(self, data):
             del data.attrs["AQUA_regridded"]
         return data
 
+    def _apply_regrid(self, data):
+        """Apply regridding if requested and supported."""
+        # We could have None data after time statistics, so we check before applying regrid.
+        if data is None:
```
oloapinivad, line 568: in which sense "none data"? I mean, it means something did not work as it should, no?

mnurisso: Yes, it is the fallback of the previous case in which if there was no time after the `Timstat` there was a `return`. 
Now having methods I cannot return, I'm not in the main method, so I return a none data after `Timstat`. If at the end of the complete evaluation chain data is `None` then I return from the main method, same behavior as before.

aqua/core/drop/drop.py
```diff
@@ -554,6 +562,39 @@ def _remove_regridded(self, data):
...
+            self.logger.warning("No data to regrid, skipping regridding step...")
+            return None
+        if self.resolution and self.resolution != "native":
+            data = self.reader.regrid(data)
+            data = self._remove_regridded(data)
+        return data
+
+    def _apply_time_stat(self, data):
```
oloapinivad, line 576: not sure if "apply" is required in all this methods names

oloapinivad (commented):
The checks on empty data are somehow puzzling to me, is only the timstat call that can lead to a collapse of the data, or am I missing something?

aqua/core/drop/drop.py
mnurisso: What do you mean by required? For me makes sense because it clarify that are all methods on the data, so that I know they are part of that subgroup, since there are many methods already

mnurisso:
> The checks on empty data are somehow puzzling to me, is only the timstat call that can lead to a collapse of the data, or am I missing something?

You don't miss anything, but if before Timstat was the last one so a failure of Timstat was leading to an immediate return now by having it in a further method means that Timstat return does not return away from the chunk computation. So if I have no time anymore after the Timstat I go anyway to the regrid. Returning a None means I can keep skipping methods as before, resulting in a return at the end of the main method. The behavior of the method itself is the same, modified internally by the decision of splitting in submethods.

aqua/core/drop/drop.py
oloapinivad: yes I was just meaning that `_regrid` `_region` etc would be enough. `apply`came from pandas and might generate confusion

aqua/core/drop/drop.py
```diff
@@ -554,6 +562,39 @@ def _remove_regridded(self, data):
...
+                exclude_incomplete=self.exclude_incomplete,
+                func_kwargs=self.stat_kwargs,
+            )
+            # data could be empty after time statistics if everything was excluded
+            if "time" in data.coords and len(data.time) == 0:
+                self.logger.warning("No data available after time statistics, skipping...")
+                return None
+        return data
+
+    def _apply_region(self, data):
```
oloapinivad, line 592: why there is no check for none data here? if the regrid is returnin none also the region would feel the same? I am wondering if having an helper method for this check would help, your are doing the same check several times

oloapinivad (approved):
Now I see the full picture. I guess you need to add a return none also for the region and then you are set.

mnurisso:
> Now I see the full picture. I guess you need to add a return none also for the region and then you are set.

perfect, done

PR #2900, opened 2026-06-09
Prettify contour line levels in plot_single_map_diff

### LINKED ISSUE #2676: [Graphics] Improve contour level selection plotting in `plot_single_map`
### Feature request

Currently contour lines in contour plots for `plot_single_map` is defining levels in a quite random manner, which is mostly based on floating points which are not really helpful and graphically pleasant. It would be nice to have a more rational method method which use integer or another "prettified" notation to avoid an excess of numbers on the contour lines

silviacaprioli:
## PR description:

Contour levels are now generated using a "nice number" algorithm (multiples of 1, 2, 2.5, 5, 10 scaled to the appropriate magnitude) instead of raw linspace values. Labels are formatted consistently with the step size, using integers when possible and scientific notation only for extreme values.

From: 
Screenshot 2026-06-09 at 18 18 11

To:
Screenshot 2026-06-09 at 19 00 38

Close #2676 

----

 - [ ] Changelog is updated.

### DIFF
--- a/aqua/core/util/graphics.py
+++ b/aqua/core/util/graphics.py
@@ -15,7 +15,6 @@ from scipy.interpolate import griddata
 
 from aqua.core.logger import log_configure
 
-
 from .sci_util import check_coordinates
 from .string import unit_to_latex
 
--- a/aqua/core/util/graphics.py
+++ b/aqua/core/util/graphics.py
@@ -479,7 +479,7 @@ def prettify_levels(vmin: float, vmax: float, nlevels: int) -> np.ndarray:
         best_step = 10 * magnitude
 
     # Build levels aligned to zero if possible (cleaner grid)
-    start = np.ceil(vmin / best_step) * best_step
+    start = np.floor(vmin / best_step) * best_step
     levels = np.arange(start, vmax + best_step * 0.5, best_step)
 
     return np.round(levels, decimals=get_decimals(best_step))
--- a/aqua/core/util/graphics.py
+++ b/aqua/core/util/graphics.py
@@ -15,6 +15,7 @@ from scipy.interpolate import griddata
 
 from aqua.core.logger import log_configure
 
+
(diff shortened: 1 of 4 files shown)


aqua/core/util/graphics.py
```diff
@@ -445,6 +445,53 @@ def apply_circular_window(ax, extent=None, apply_black_circle=False):
...
+    raw_step = data_range / nlevels
+
+    # Candidate "nice" multipliers
+    nice_multipliers = [1, 2, 2.5, 5, 10]
+    magnitude = 10 ** np.floor(np.log10(abs(raw_step)))
+
+    best_step = None
+    for m in nice_multipliers:
+        candidate = m * magnitude
+        if candidate >= raw_step:
```
oloapinivad, line 474: I have some doubt here, chosing the best_step as larger than the the raw_step. Wouldn't be better to find the one which is closer to it?

aqua/core/util/graphics.py
```diff
@@ -445,6 +445,53 @@ def apply_circular_window(ax, extent=None, apply_black_circle=False):
...
+            best_step = candidate
+            break
+
+    if best_step is None:
+        best_step = 10 * magnitude
+
+    # Build levels aligned to zero if possible (cleaner grid)
+    start = np.ceil(vmin / best_step) * best_step
```
oloapinivad: uhm, this is doing ceiling on lowest number, wouldn't this exlucde lowest values? Perhaps it is not a problem since this done only for contours, correct? 

```
start = np.floor(vmin / best_step) * best_step
```
however, in theory we can use the same for shading...

aqua/core/util/graphics.py
silviacaprioli: I tried as suggested but I end up with maybe too many levels...
image

tovaz92 (approved):

PR #2901, opened 2026-06-10
Drop level

### LINKED ISSUE #1878: Level selection for LRA generation
### Feature request

It would be ideal, for some applications, to have the possibility to generate the LRA for a specifi level for 3D fields.

mnurisso:
## PR description:

Adding level selection to the DROP capabilities.
This is done with a target field in the target block of the yaml file or the --level flag.
The level can be a single value (int or float) or a list. Filename and selection is handled accordingly for the two cases.

Tests are added and level selection tests are expanded.

Close #1878 

## People involved:

@mcadau @oloapinivad 

 - [x] Changelog is updated.


mnurisso:
This is ready to be reviewed, the only test failing is the MN5 bridge reading, which is today under maintainance

aqua/core/console/drop.py
```diff
@@ -161,6 +163,10 @@ def drop_execute(args):
     regrid_first = get_arg(args, "regrid_first", _cfg(config, "target", "regrid_first", False))
     startdate = get_arg(args, "startdate", _cfg(config, "target", "startdate"))
     enddate = get_arg(args, "enddate", _cfg(config, "target", "enddate"))
+    level = get_arg(args, "level", _cfg(config, "target", "level"))
+    # Convert comma-separated level string to list if necessary
+    if isinstance(level, str) and level:
+        level = [float(lev) if "." in lev else int(lev) for lev in level.split(",")]
```
oloapinivad, line 169: I am not sure I understad why you use `if "."` ? I assume people will pass `850` which does not include any dots, so perhaps something as `isdigit()` would be better to avoid the strsplit?

aqua/core/drop/drop_writer_base.py
```diff
@@ -258,7 +258,7 @@ def _iter_months_in_year(self, year_data, performance_reporting):
         for month in months:
             yield month, year_data.sel(time=year_data.time.dt.month == month)
 
-    def _write_chunk(self, data, var, year, month, dask=False, performance_reporting=False):
+    def _write_chunk(self, data, var, year, month, level=None, dask=False, performance_reporting=False):
```
oloapinivad, line 261: for future memory, I am starting to wonder if the filename generation can lie externally from the chunk and passed as an argument only, this would avoid passing all those arguments

aqua/core/drop/output_path_builder.py
```diff
@@ -74,7 +72,7 @@ def build_path(self, basedir, var, year=None, month=None, day=None, output_forma
             str: The full path to the output file.
         """
         folder = self.build_directory()
-        filename = self.build_filename(var, year, month, day, output_format=output_format)
+        filename = self.build_filename(var, level, year, month, day, output_format=output_format)
```
oloapinivad, line 75: I am wondering this should be extracted to higher level

oloapinivad (approved):
All good, just a question on the first parsing of the string for level. there are a lot of nested calls, we can at least discuss briefly what could be done to improve it without actually implementing it

mcadau (approved):
After a successful LRA generation with level selection, this can be merged I guess. Maybe you can just take a look on the tests (it seems there's something wrong) and, once tests are passing, feel free to merge

aqua/core/drop/drop.py
mcadau: Should the level be specified here as well?

mnurisso: Done

mnurisso:
With many hiccups from data-porfolio and similar tests are passing. Merging.

PR #2910, opened 2026-06-19
More info in last_update key for content.yaml

jhardenberg:
## PR description:

This small change adds more info on the origin of a push to lumi-o by including in the last_update key of  `content.yaml` the username, the hostname and the ip of the machine on which the file was created.
This will help to track responsibility for the latest pushes to lumi-o, helping in case of debugging.
It allows in theory to drop completely the update of the updated.txt file on the aqua-web repo on github which was kept just for this purpose.

NB: there is already an existing option `(-d, --no-update)` which prevents update on github. If we wish we could either make this behaviour default or simply ask the workflow to always use the `-d` option if we think that this can be dropped.

----

 - [ ] Changelog is updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 Unreleased in the current development version (target v1.0.0):
 
 Complete list:
+- More info on the origin of a push to lumi-o in the logs (#2910)
 - DROP: `drop` option from the `region` block is correctly handled, specifying if the NaN around the region should be kept or not (#2903)
 - DROP: add a `regrid_first` option to perform regridding before time statistics, useful when time-statistics can remove spatial coords (#2899)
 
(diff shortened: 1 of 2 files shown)


jhardenberg:
Actually @mnurisso I see that  the workflow started using the `--no-update` option already two weeks ago https://gitlab.earth.bsc.es/digital-twins/de_340-3/workflow/-/merge_requests/1332 . Since a secondary function of that update of aqua-web (which I admit was creating a dependency on a now deprecated repository - so that is good)  was also to keep a record of who last updated lumi-o, the present solution recovers that  information by storing it directly on lumi-o.

jhardenberg:
PS: if we should anyway make a release for operational, it would not be bad to include this also there (could be useful in case we need to figure out easily where some particular data pushed to lumi-o come from)

mcadau (approved):

mcadau:
This can be helpful indeed. Is changelog missing? In any case, we can include this also before the new operational release

silviacaprioli:
That's good but we need to open another PR to point to v019

PR #2917, opened 2026-06-22
Revert "config to 2025 (#2855)"

silviacaprioli:
This reverts commit aada3c0a6e18e0b571095980fb28cc1228022f39.

We realized that data on DVC have not been updated yet (and descriptions on the DESP are inaccurate if there are less data then those indicated on the config files).


jhardenberg (approved):
as discussed

PR #2923, opened 2026-06-28
aqua analysis logger bug

jhardenberg:
## PR description:

This fixes a small `aqua analysis` bug: if `nmaxprocesses` is not specified logging fails because its value is None.


jhardenberg:
so small, that I would avoid a changelog entry

jhardenberg:
#2897 should know about this

silviacaprioli (approved):

PR #2931, opened 2026-06-30
Implementation of polytope in the refactored reader

mnurisso:
## PR description:

Working example:

GSV Polytope access
```python
reader = Reader(
    catalog='climatedt-gen2',
    model='ICON-5km',
    exp='baseline-hist',
    source='monthly-hpz7-sfc',
    engine='polytope'
)

```


mnurisso:
With the current implementation var selection does not work.
`tos` is present in the final dataset, but `var='tos'` does not find any message. I presume the old loop implementation is still needed.

mnurisso:
Fixed

PR #2935, opened 2026-07-02
[Backend] Remove intake-xarray dependency

mcadau:
## PR description:
With this PR, the aqua-backed refactor is completed removing the `intake-xarray` dependency.
Work in progress.

----


mcadau:
I changed the structure, this time explicitly having the readers and the openers. `compat.py` seems still mandatory at least before a Climate-DT-catalog PR to remove intake-xarray entries in the catalog yaml files. 
For the rest, the structure is identical to IceChunk when appliable. 
Test are also passing

mnurisso:
> I changed the structure, this time explicitly having the readers and the openers. `compat.py` seems still mandatory at least before a Climate-DT-catalog PR to remove intake-xarray entries in the catalog yaml files. For the rest, the structure is identical to IceChunk when appliable. Test are also passing

I'm sorry to be picky on this but this PR is still disruptive on other drivers and in my opinion I wouldn't merge it.
The  IntakeSourceAdapter(base.DataSource) is a step back towards intake 1 and you're applying it to all the drivers, bringing them one step back.
The opener and reader structure which is now implemented is in the good direction but I'd start on a fresh branch so that you don't modify other drivers structure.

aqua/core/fixer/fixer.py
```diff
@@ -231,12 +231,6 @@ def fixer(self, data, destvar, apply_unit_fix=True):
 
                 # adjust units
                 if tgt_units:
-                    self.logger.warning(
-                        "Variable %s has fixer target units %s, but source units are %s",
```
jhardenberg, line 235: Why has this been removed?

aqua/core/backend/catalog_mixin.py
```diff
@@ -67,8 +67,9 @@ def setup_catalog(
         self.kwargs = self._format_realization_reader_kwargs(self.kwargs)
         self.logger.debug("Using filtered kwargs: %s", self.kwargs)
 
-        # HACK for intake2 following https://github.com/intake/intake-xarray/issues/150
-        # Rebuild esmcat with the filtered (and possibly realization-formatted) kwargs.
+        # Rebuild esmcat with the filtered (and possibly realization-formatted) kwargs:
```
jhardenberg, line 70: Why change the comment?

aqua/core/intake_drivers/fdb/base.py
```diff
@@ -1,22 +1,15 @@
-from intake.source import base
+"""FDB source adapter built on the shared AQUA intake 2 adapter."""
```
jhardenberg, line 1: There should be no need to change base.py. It works perfectly well with gsv and z3fdb .....

aqua/core/intake_drivers/fdb/base.py
```diff
@@ -1,22 +1,15 @@
...
 
+from ..base import IntakeSourceAdapter
```
jhardenberg, line 3: from ..base ? We are in base here

aqua/core/intake_drivers/icechunk/icechunk.py
```diff
@@ -1,10 +1,9 @@
-from intake.source import base
-
+from ..base import IntakeSourceAdapter
```
jhardenberg, line 1: where is the base this is importing from? 
Again, why did we need to change this?

aqua/core/intake_drivers/xarray/openers/xarray.py
```diff
@@ -0,0 +1,143 @@
+"""Functional entry points for NetCDF and Zarr access through xarray.
```
jhardenberg, line 1: Why do we need all this? there is already a XArrayDatasetReader() in intake/readers/readers.py

jhardenberg:
Frankly, all this seems overbloated compared to what we need. The goal of having our own intake_xarray driver is simply to be able to make small modifications/fixes without depending on intake-xarray. I would start from a clone of intake-xarray and modify it though .... here also functionality already present in intake is rewritten (which may not be necessary)

jhardenberg:
I would start a fresh branch/PR with a clone of intake-xarray and then just add the minimum modifications that we need (for example in the backend there is still a hack to expose the following:

        # HACK: convenience to get expanded url, xarray_kwargs and metadata for netcdf/zarr sources for intake2.
        # This provides direct access to the intake data object and is xarray-specific.
        self.esmcat.data = self.esmcat.reader.kwargs["args"][0]
        self.esmcat.metadata = self.esmcat.reader.metadata
        self.esmcat.xarray_kwargs = self.esmcat._entry._captured_init_kwargs.get("args", {}).get("xarray_kwargs", {})
        
 This would not be needed if intake-xarray remembered to expose these.

PR #2938, opened 2026-07-02
Catalog generator based on STAC for ClimateDT official runs

### LINKED ISSUE #2941: Intake catalog generator with STAC query
### Feature request

For ClimateDT operational data we are developing a tool able to generate a full set of catalog by interrogating the STAC catalog provided by ECMWF https://qubed.lumi.apps.dte.destination-earth.eu/
The implementation is done in #2938 

Currently, we are missing a few things
- [x] A method to detect model resolution: from the STAC point of view, and from the MARS requet therein, the original resolution of the model as well as the output data is impossible to discover. we need to find a correct way to do that which is not trivial
- [x] Similarly, we have no way to assign the right vertical levels to the oceanic coordinates. Currently this is deduced from the data-portfolio, which requires a clone of the catalog and knowing the tag used (not trivial)
- [x] A CLI implementation, similarly to `aqua catgen` is  required
- [x] We need to develop tests and compare against the current catalog
- [x] Once all is set, a Github action into the ClimateDT catalog can be implement to achieve the automatic update of the catalog itself. 

Some further clarification can be obtained by interacting with ECMWF: 
- we need to understand where monthly data is, since apparently is not available on the STAC
- we need to have the access for MN5 STAC catalog to get the catalog for data on MN5

oloapinivad:
## PR description:

This implements the idea of using the STAC API https://qubed.lumi.apps.dte.destination-earth.eu/ to automatically produce the aqua intake catalogs for climatedt-gen2 experiments. 

This can be tested with the following code, it tooks a couple of minutes to run and produce a complete catalog for all the experiments available on Lumi (not on MN5). A series of small hack have been introduced and can be refined later, but it can be a starting point to then have an automatic action that runs on the ClimateDT catalog to always have a version of the bridge available data. 

```python
from aqua.core.catgen.stacgen import AquaSTACGenerator
generator = AquaSTACGenerator(
    catalog="climate-dt-gen2",
    loglevel="debug"
)
generator.explore_tree(
    layers = [
        "activity", "experiment", "model", "realization",
        "expver", "stream", "resolution", "levtype"
    ],
    filters={
        "activity": ["baseline", "projections"],
        "model": ["icon", "ifs-fesom"],
        "expver": "0001",
        "stream": ["clte"],
    }
)
generator.complete_request()
generator.generate_catalog(catalog_dir_path="/Users/paolo/Desktop/Climate-DT-catalog/", catalog_name="climatedt-gen2-stac")
```

It reuses some of the features, as the jinja file and the configuratiom matchin grids fro the default catgen. 

 - [ ] Changelog is updated.


oloapinivad:
I made some upgrade:
- stacgen CLI 
- hardcoded schema to guess model resolution
- scraping from the web of the levels for the ocean (horrible but works!)

I am doing some verification but it seems to be working quite well

oloapinivad:
I incidentaly added the automatic grid deployer feature.

aqua/core/console/main.py
```diff
@@ -18,7 +18,10 @@
 
 # this are used to check existence of aqua.diagnostics
 from .install import DIAGNOSTIC_CONFIG_DIRECTORIES, DIAGNOSTIC_TEMPLATE_DIRECTORIES, InstallMixin
+
+# parser
```
mnurisso, line 22: why this line?

aqua/core/catgen/stacgen.py
```diff
@@ -0,0 +1,882 @@
...
+            str: Grid identifier (e.g. "nemo-eORCA12-hpz7-nested-v3").
+
+        Raises:
+            KeyError: If model or levtype not found in matching_grids.yaml.
+        """
+        grids = self._load_grids_config()
+
+        # Determine tier: production for baseline/projection, develop otherwise
+        activity = request.get("activity", "").lower()
+        tier = "production" if activity in ("baseline", "projection") else "develop"
```
mnurisso, line 440: it should be projections, not singular

aqua/core/catgen/stacgen.py
```diff
@@ -0,0 +1,882 @@
...
+
+        Args:
+            *pairs: Variable number of (key, value) tuples.
+
+        Returns:
+            CatalogNode: The leaf node of the added chain.
+        """
+        node = self
+        for key, value in pairs:
+            if value in node.children:
```
mnurisso, line 219: Partially from AI check:

Since you are checking by `child.value` and not by `(child.key, child.value)` you may collaps in the same node which are not supposed to be collapsed, for example:

```
("activity", "baseline")
("experiment", "baseline")
```

This is not an issue now, but we may have values which are the same for different keys in the future and I'd make this method stronger.

aqua/core/catgen/stacgen.py
```diff
@@ -0,0 +1,882 @@
...
+
+        Args:
+            response: STAC API response JSON.
+            key: Facet key name to extract (e.g. "experiment", "model").
+
+        Returns:
+            list: Enum values for the key, or empty list if not found.
+        """
+        for element in response.get("links", []):
+            if element["title"] == key:
```
mnurisso, line 585: Probably not an issue but there is no fallback here if title is not there

mnurisso (commented):
I left some general comment to be addressed before a merge. Additionally a proper test should be written in order to have enough coverage.

In the long run probably to move in the climatedt specific repo

PR #2940, opened 2026-07-03
[Backend] Aqua backend z3fdb

jhardenberg:
## PR description:

Implement z3fdb intake driver

 - [x] Changelog is updated.


jhardenberg:
a sample z3fdb config file is in aqua/core/config   at the moment. Any better suggestion on where to place it would be welcome.

Test with:

```
from aqua import Reader
reader = Reader(model="IFS-FESOM-5km", exp="baseline-hist", source="hourly-hpz7-sfc", catalog="climatedt-gen2", engine='z3fdb', loglevel='debug', config_z3fdb="/home/jvonhar/.aqua/config-z3fdb.yaml")
data = reader.retrieve(var='2t')
data
```

If not `config_z3fdb` key is specified, it will look in the current directory for a file named `config_z3fdb.yaml`

jhardenberg:
Currently it still returns `var167` if I ask for `2t`: there is something to be fixed in the fixer

jhardenberg:
For some reason it adds `AQUA_config_z3fdb :
/home/jvonhar/work/AQUA/config-z3fdb.yaml`to the variable attributes .... @mnurisso any idea why?

jhardenberg:
> Currently it still returns `var167` if I ask for `2t`: there is something to be fixed in the fixer

I added a small fix to the fixer to allow the fixer to work correctly also if grib code variables (eg var167) are passed.

I also extended slightly our conventions file to support 10si and tcw which were missing. Since it is something which is not urgent for the main I include it here.

jhardenberg:
The config file is now in ~/.aqua and I added it to the cli install and z3fdb driver should find it automatically base don the AQUA ConfigPath. So it is not necessary anymore to specify it. @mnurisso please check if all looks good to you.

jhardenberg:
chunking support is still missing

jhardenberg:
It turns out that z3fdb supports only two chunking options: SINGLEVALUE (each step or level) or NONE. 

chunking is now supported this way:
- chunking in the time direction is always SINGLEVALUE 
- chunking in the level direction is by default NONE. If `chunks`is defined and it is a dictionary with a 'level' key then chunking SINGLEVALUE is done also in the level direction. So `chunks={"level": 1}` provides single-level chunking (and anyway time chunking).  If something like `chunks={"level": 3}` is passed then lthe 3 is ignored and evel chunking is still single-value.

Please let me know if you agree with this solution.

jhardenberg:
Now the resulting vertical coordinate has correct units and values if specified in the catalog.

jhardenberg:
It turns out that the config file to provide to z3fdb is a normal fdb config file. this opens the possibility to write tests.

aqua/core/console/install.py
```diff
@@ -293,6 +293,11 @@ def install(self, args):
                 os.path.join(self.corepath, "config-aqua.tmpl"), os.path.join(self.configpath, "config-aqua.yaml")
             )
 
+            self.logger.debug("Creating config-z3fdb.yaml configuration file")
```
oloapinivad: why are we installing the config from z3fdb? I mean this is practical, but it is a very specific use case...

aqua/core/fixer/fixer.py
```diff
@@ -149,7 +149,10 @@ def fixer(self, data, destvar, apply_unit_fix=True):
                 source = to_list(varfix.get("source", None))
                 # We want to process a list of sources
                 if source:
-                    match = list(set(source) & set(data.variables))
+                    source_strs = {str(s) for s in source}
```
oloapinivad, line 153: what is the goal of this block? you are trying to make a match if the strings are codes? Shall we use the `or` instead of the `|`?

aqua/core/fixer/fixer.py
```diff
@@ -221,6 +219,10 @@ def fixer(self, data, destvar, apply_unit_fix=True):
                         # Already adjust all attributes but not yet units
                         if att == "units":
                             tgt_units = value
+                            # If units are not present and we know what they should be, let's set them
+                            if "units" not in data[source].attrs:
+                                data[source].attrs["units"] = value
```
oloapinivad: isn't this assuming that we have a `value` for units?

aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py
```diff
@@ -0,0 +1,345 @@
...
+                years = range(int(d0[:4]), int(d1[:4]) + 1)
+
+        iso_start = f"{d0[:4]}-{d0[4:6]}-{d0[6:8]}"
+    else:
+        ys = list(years)
+        d0 = f"{ys[0]}0101"
+        d1 = f"{ys[-1]}1231"
+        iso_start = f"{ys[0]}-01-01"
+
+    if freq == "h":
```
oloapinivad: such if then else is structure might be similar to something available in the time utils already, worth double checking to not write a new function to do the same thing

aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py
```diff
@@ -0,0 +1,345 @@
...
+            "lat_bounds": (("cell", "vertices"), lat_b_vals),
+        }
+    )
+    return ds
+
+
+def add_coordinates(ds, levunits=None):
+    """Add coordinates based on grid type."""
+
+    if "cell" in ds.dims:
```
oloapinivad: please comment this clearly (even with a HACK), this is a strong assumption based on the idea that any unstructured grid coming from FDB is healpix

aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py
```diff
@@ -0,0 +1,345 @@
...
+        is_healpix = True
+    else:
+        is_healpix = False
+
+    if is_healpix:
+        ncell = ds.sizes["cell"]
+        is_healpix = (
+            ncell % 12 == 0 and (ncell // 12).bit_length() - 1 == np.log2(ncell // 12) and (ncell // 12).bit_length() % 2 == 1
```
oloapinivad: is there as any astropy_healpix to do this?

aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py
```diff
@@ -0,0 +1,345 @@
...
+    if not isinstance(variables, (list, tuple, range)):
+        variables = [variables]
+
+    if levels:
+        request["levelist"] = levels
+    else:
+        levels = request.get("levelist", None)
+
+    # Set level units based on level type
+    levtype = request.get("levtype", None).lower()
```
oloapinivad: but we do not have any unit for the 3d data? or can't we assign them with the fixer instead of hardcoding them here?

aqua/core/intake_drivers/fdb/fdb.py
```diff
@@ -40,7 +40,7 @@ def __init__(
         # IntakeFDBSource.instancecount += 1
         # print("Number of FDB source calls = " + str(IntakeFDBSource.instancecount))
 
-        if engine == "polytope":
+        if engine == "polytope-new":
```
oloapinivad: what the `new` means?

docs/sphinx/source/adding_data.rst
```diff
@@ -233,6 +233,12 @@ Some of the parameters are here described:
     This optional parameter is used to specify the expver of the data on the HPC FDB.
     If not set, the expver is assumed to be the same for all data.
 
+.. option:: engine
+
+    This optional parameter is used to specify the data retrieval engine.
+    The default is ``fdb``. Alternatives are ``polytope`` (see :ref:`polytope`) and ``z3fdb``.
```
oloapinivad: isn't the if on `polytope-new` now?

docs/sphinx/source/adding_data.rst
```diff
@@ -233,6 +233,12 @@ Some of the parameters are here described:
...
+    To use the ``z3fdb`` alternative access method, you must specify ``engine="z3fdb"`` when instantiating the ``Reader`` class, or configure it as ``engine: z3fdb`` in the intake catalog source entry.
```
oloapinivad: I think that we should also pass a `config_z3fdb` file, correct?

docs/sphinx/source/polytope.rst
```diff
@@ -53,6 +53,33 @@ Lumi Databridge is the default endpoint, but you can specify the MN5 Databridge
...
+    data = reader.retrieve(var='2t')
+
+For this engine, a configuration file ``config-z3fdb.yaml`` is used (which is copied to your configuration folder during installation).
+
+Chunking Logic for z3fdb
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+When using ``engine="z3fdb"``, chunking is configured as follows:
+
+* **Time direction**: Chunking is always ``SINGLEVALUE``.
```
oloapinivad: isn't this duplicated from above?

aqua/core/console/install.py
mnurisso: I would add an HACK comment just to be aware of it

jhardenberg: In fact I am considering to remove it. It turns out that the config file required is a standard fdb config file. The plan agreed with @mnurisso is to merge this one more or less as it is (after considering your following comments) and then immediately prepare a new branch where we unify further the gsv driver and the z3fdb driver . The config file kwarg `config_z3fdb` will be renamed `config_fdb` and will work also to override fdb_home_bridge of the gsv driver

aqua/core/fixer/fixer.py
jhardenberg: This fixes an issue which we had previously in the fixer: we need to make sure that even if 130 is passed as a source it is a string and not a digit. Further, our convention files mention grib codes directly as digits, but we may also have "var130" (instead of 130) as var names. So the following line includes also the "varXXX" versions among the sources. "|" is the "set union" operator, it cannot be substituted by or

jhardenberg: I added a comment to explain it

aqua/core/fixer/fixer.py
jhardenberg: we are iterating over attributes. If it has "units" it will have a value

aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py
jhardenberg: Indeed. My plan is to merge this one as it is (so @mnurisso can start to use it) but immediately make a new branch to refactor and unify more z3fdb and gsv drivers where possible

aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py
```diff
@@ -0,0 +1,345 @@
...
+    req = {}
+
+    if years is None and (start_date is None or end_date is None):
+        years = request["year"]
+        if not isinstance(years, (list, tuple, range)):
+            years = [years]
+        years = [str(y) for y in years]
+
+    if start_date is not None or end_date is not None:
+        if start_date is None or end_date is None:
```
mnurisso, line 101: It would be nice to have the possibility as the other tools to read from the source intake the end_date or the start_date in the available ones. Is this a feature that makes sense for this opener?

aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py
mnurisso: It is more probably worth to go the other way, using this and deprecating our block

aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py
```diff
@@ -0,0 +1,345 @@
...
+    variables=None,
+    levels=None,
+):
+    """Wrap a z3fdb zarr array as an xarray.Dataset with one variable per param.
+
+    Zarr shape is (time, param, cell) or, when levels is given,
+    (time, param, level, cell). Each param slice becomes its own
+    DataArray; dask keeps everything lazy.
+
+    Parameters
```
mnurisso: To be converted as google style docstrings

aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py
mnurisso: cannot be added a double check on the number of cells and if not compliant raise a `NotImplementedError`?

aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py
```diff
@@ -0,0 +1,345 @@
...
+    else:
+        is_healpix = False
+
+    if is_healpix:
+        ncell = ds.sizes["cell"]
+        is_healpix = (
+            ncell % 12 == 0 and (ncell // 12).bit_length() - 1 == np.log2(ncell // 12) and (ncell // 12).bit_length() % 2 == 1
+        )
+
+    if not is_healpix:
```
mnurisso: I think referring to my comment and paolo above this is what at least I was asking for

mnurisso (approved):
I'm going to test it on desp. I left some comments together with the ones by Paolo to be checked but it seems ok to me

aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py
jhardenberg: unfortunately the `isnpixok()` function is not available in astropy_healpix. We rewrote it in `aqua.core.util.graphics` but it is a simple formula, so in order not to create a not needed dependency I would repeat it here.

aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py
jhardenberg: It is better to assign them here because then the datamodel will understand that this is a vertical axis and fix all the rest automatically.  The units of 3D data are implicit in the `levtype` key. Also GSV_interface hardcodes them: https://gitlab.earth.bsc.es/digital-twins/de_340-3/gsv_interface/-/blob/main/gsv/levels/pressure_level_reader.py

aqua/core/intake_drivers/fdb/fdb.py
jhardenberg: It was temporary, thanks. Anyway the idea is that even if at a certain point we write a native polytope reader, it should still be possible to call the gsv_interface version which also supports polytope. I now renamed this so that if `engine="polytope"` is passed, the new driver is called (which at the moment is still doing the same) and if "polytope-gsv" is passed the gsv-enabled version is used

aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py
jhardenberg: to be explored, yes

aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py
jhardenberg: that is exactly what we do below. I switched to `NotImplementedError` as suggested

docs/sphinx/source/adding_data.rst
jhardenberg: I renamed to "polytope" and "polytope-gsv" as discussed above  and updated the docs

docs/sphinx/source/adding_data.rst
jhardenberg: Not compulsory at the moment. If none is passed it finds the one in `.aqua`. Anyway I will change this again in the next iteration. Probably we will not distribute that file anymore with AQUA.

docs/sphinx/source/polytope.rst
jhardenberg: The note in `adding_data.rst` contains the same info but it is where we explain the API.  It is probably good to repeat it here. I rephrased a bit to make it clearer.

aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py
jhardenberg: It is already being done at line 314. When it arrives here that has already been done.

aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py
jhardenberg: ok, done

aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py
jhardenberg: yes, this is already there

jhardenberg:
the kwarg was renamed config_fdb

aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py
oloapinivad: I realized the same while playing for smmregrid.

mnurisso:
Last commit totally breaks the DESP functionality. Testing the commit before I can confirm that dask now works.
I'd like to revert the last commit or have a clear explanation on how to work not on lumi and possibly give with the catalog or here the config file. I cannot guess how it should be written the fdb home for this driver, it should work out of the box for the end user.

jhardenberg:
The driver has been refactored with:
- fully flexible time axis (now any date/time period is possible)
- config_fdb keyword (not needed but useful)
- works with HPC or bridge
Tests to be added

jhardenberg:
The z3fdb driver should now be a full substitute for the gsv driver and it has the same capabilities (but much faster!)
Mixed reading from hpc and bridge has been added (should be tested in practice: any catalog to try?)
Tests are now included.  z3fdb is run with most gsv tests + specific ones and we are trying to reach sufficient coverage.

oloapinivad:
That's super cool!
But tests are broken, any idea why?

jhardenberg:
> That's super cool! But tests are broken, any idea why?

not my tests.... it is the console test which is complaining. 
I was also looking forward to the tests because I am aiming at a green mark on covergae....

jhardenberg:
ok, tests look good and coverage has increased

jhardenberg:
Do not delete branch for now (request by @mnurisso )

PR #2944, opened 2026-07-06
[Backend] introduce select minimum sample

### LINKED ISSUE #2943: [Backend] Improve `_retrieve_plain` in the backend
### Feature request

The new implementation of `_retrieve_plain` in the backend refactoring #2926 is still to be refined, despite providing a more flexible solution
For the xarray data source:
- the variable selection is still missing based on the gridtype and for this we are waiting a new version of `smmregrid`
- The chunking is still automatic or even lacking in some cases, we need to find a way to pass at least on time dimension the most convenient chunking possible when loading the data. This is limited by the factor that we do not know the time dimension name before loading.

oloapinivad:
## PR description:

This PR simplifies and upgrade the behaviour of the backend `_retrieve_plain` by using the new features from the `smmregrid==0.1.16`. It extract the minimum sample of data available on a dataset based on gridtypes.


oloapinivad:
This now is working with the new release of smmregrid. it allows for 
1. reading the startdate/enddate from metadata to allow for `retrieve_plain` from a specified point
2. supports the cds api set up for https://cds.climate.copernicus.eu/datasets/reanalysis-era5-single-levels?tab=analysis_ready_data
3. introduces a check to verify that data loaded by the reader is actually dask array

We are missing

- [ ] tests and documentation for the cdsapi
- [x] a centralization of the way that reader_kwargs and metadata in general are configured. it would be nice to have everything inside the `CatalogMixin`

oloapinivad:
I would merge this one into the backend too

PR #2950, opened 2026-07-09
[Backend] use `Coordsidentifier` to detect vertical coordinate

oloapinivad:
## PR description:

This removes the `find_vert_coord` function in favour of the CoordsIdentifier class  from the datamodel.
It is not very neat but it should do the job


oloapinivad:
This looks ready to be merged

PR #2957, opened 2026-07-13
[DataModel] match coordiantes of area and the data

maqsoodrajput:
When areas and data lon were disagreeing `fldmean` was failing. 
Now, it works if the data is in `[-180,180]` or `[0, 360]`. 
2t_mean_2


maqsoodrajput:
ruff format remaining

oloapinivad:
Thanks Maqsood, this is very interesting and I feel it is a good add on. I will review it asap.
Can you share an example when this is happening providing a sample data (one record)? 
I am wondering when this might happen, because the if the area file is generated from the data this should not be a problem.... it might only happen when the grid file is misaligned with data we are using.

aqua/core/fldstat/fldstat.py
```diff
@@ -338,12 +338,23 @@ def align_area_coordinates(self, data: xr.Dataset | xr.DataArray, decimals: int
...
+                # Tolerance-based check. This replaces a fragile round-then-compare
+                # approach: on grids whose spacing is a power-of-two fraction of 360
+                # (e.g. reduced Gaussian grids like tco1279), many coordinate values
+                # land exactly on a decimal rounding boundary (e.g. ...875 at 5
+                # decimals). Two independently-computed float64 representations of
+                # the "same" value can then round to different neighbours purely
+                # from sub-1e-6 representation noise, even though they are
+                # physically identical. np.isclose with an absolute+relative
+                # tolerance avoids this instability.
+                atol = 1.5 * 10 ** (-decimals)
```
oloapinivad: I would make the tolerance as a constant to be declared at the beginning of the module

oloapinivad:
So my main concern is that the `CoordTransformer` is now doing a job that was already done by the `CoordIdentifier`. 
Datamodel has already a method to discover if coordinates is [0,360] or [-180,180]. We inappropriately call it "convention" (this can be fixed), and can be `centered` or `positive` (we can also use more precise name here). See the code below 

https://github.com/DestinE-Climate-DT/AQUA/blob/fadf24329bf7bf81bb23ef9589bf2289847ec178/aqua/core/data_model/coordidentifier.py#L340-L354

This means that `CoordTransformer` should only verify if the `convention` is not the one defined in our aqua datamodel, and then apply the conversion. The core structure of your PR is robust, I would only simplify this so that we do not need to it twice.

aqua/core/config/data_model/aqua.yaml
```diff
@@ -19,6 +19,7 @@ data_model:
     long_name: longitude
     stored_direction: increasing
     units: degrees_east
+    range: [0, 360]
```
oloapinivad: I think that range is a good alternative to `convention`

oloapinivad: about the values to use, a literal is safer because it can be written into a netcdf files as an attribute while a list not. so perhaps staying with `centered` and `positive` as in the current solution is safer. Alternatively, we can use `0_360` and `180_180`

aqua/core/data_model/coordtransformer.py
```diff
@@ -286,12 +287,80 @@ def _convert_bounds(self, data, src_coord, tgt_coord, factor):
...
+            data (xr.Dataset or xr.DataArray): The Xarray object.
+            tgt_coord (dict): Target coordinate dictionary, as defined in the
+                data model YAML. Only acts on the longitude coordinate, and only
+                if a "range": [lo, hi] key is present; otherwise a no-op so this
+                stays fully backward compatible with data models that don't set it.
+
+        Returns:
+            xr.Dataset or xr.DataArray: The Xarray object with longitude wrapped.
+        """
+        coord_name = tgt_coord.get("name")
```
oloapinivad: this is probably an overkill, since CoordTransformer is running on top of CoordIdentifier, which means that if you have a `lon` you should have found it. 

An example of an identification of `longitude`

```python
 'longitude': {'name': 'lon',
  'dims': ('lon',),
  'units': 'degrees_east',
  'range': (np.float64(0.0), np.float64(340.0)),
  'bounds': None,
  'stored_direction': 'increasing',
  'convention': 'positive',
  'confidence_score': 300,
  'matched_attributes': ['name', 'standard_name', 'axis', 'units']}
```

aqua/core/data_model/coordtransformer.py
```diff
@@ -286,12 +287,80 @@ def _convert_bounds(self, data, src_coord, tgt_coord, factor):
...
+        if coord_name not in ("lon", "longitude"):
+            return data
+        if coord_name not in data.coords:
+            return data
+
+        lon_range = tgt_coord.get("range")
```
oloapinivad: if we decided that `range` has to be declared - and I think it is a good idea - this can be removed

aqua/core/data_model/coordtransformer.py
```diff
@@ -286,12 +287,80 @@ def _convert_bounds(self, data, src_coord, tgt_coord, factor):
...
+
+        return data
+
     def assign_attributes(self, data, tgt_coord):
         """
         Assign attributes to the coordinate.
         """
         for key, value in tgt_coord.items():
-            if key not in ["name", "units", "positive", "stored_direction", "bounds"]:
+            if key not in ["name", "units", "positive", "stored_direction", "bounds", "range"]:
```
oloapinivad: we might want to declare this at the top of the module to make it clearer

tests/test_fldstat.py
```diff
@@ -347,3 +349,122 @@ def test_accessor_wrappers_match_reader(self, reader_ifs, data_ifs, method_name,
         reader_result = reader_ifs.fldstat(data_var, stat=stat_name)
 
         assert accessor_result.equals(reader_result)
+
+
+@pytest.mark.aqua
+class TestAlignAreaCoordinates:
```
oloapinivad, line 365: don't we have any real file to be used instead of creating a wide range of artificial xarray?

aqua/core/config/data_model/aqua.yaml
oloapinivad: on a second thought, we already have `range` among the attributes from `CoordsIdentifier`. also, how this can work with regional data?
I am wondering if there is any official value for CF convention

maqsoodrajput:
> Thanks Maqsood, this is very interesting and I feel it is a good add on. I will review it asap. Can you share an example when this is happening providing a sample data (one record)? I am wondering when this might happen, because the if the area file is generated from the data this should not be a problem.... it might only happen when the grid file is misaligned with data we are using.

Thanks for doing the review. Yes, so the issue was that the data and the grid lon were not aligned. 
I have copied some sample data here: `/work/ab0995/a270260/AQUA_v1/testing_scripts/test_zarr_store/IFS-FESOM_native_data.zarr`. This is IFS-FESOM tco1279.

aqua/core/config/data_model/aqua.yaml
maqsoodrajput: I agree with the `0_360` suggestion (updated the code accordingly).  

On the second comment however, I think we should keep it as a safeguard. In case if anyone introduces `range` key to `time` or `latitude`.

oloapinivad:
Hi @silviacaprioli @maqsoodrajput what is going on in this PR? It seemed it was all solved and now tests are failing again, some conflict with the main?

silviacaprioli:
I think we need to merge the main. I see the tests are failing with the error we had last week in the main branch and now is solved

oloapinivad:
@maqsoodrajput can we merge this now?

maqsoodrajput:
@oloapinivad this PR diverges from the main in the file `aqua/core/data_model/coord_utils.py` where in the main branch  we have, 
```
# Possible basic names for coordinates
from aqua.core.default import DEFAULT_COORD_NAMES
``` 
can we safely remove this?

oloapinivad:
Hi @maqsoodrajput well spotted. I move all the new `AQUA_DEFAULT` to the `defaults.py` so that they are all centralised.

oloapinivad (approved):

oloapinivad (approved):

PR #2958, opened 2026-07-14
[Backend] remove intake-xarray (2)

### LINKED ISSUE #3064: Intake-xarray reads non-dask Datasets from sources with a single file
### machine

Lumi

### catalog

ci

### version

main

### What happened?

If we read a source using intake-xarray based on a single netcdf file, like in the following example, the result is a Dataset without dask. This is probably because in these cases intake-xarray uses `open_dataset` instead of `open_mfdataset` which have different defaults on dask usage. Notice that the backend now warns us with an error (this is probably new)

```python
from aqua import Reader
reader = Reader(catalog='ci', model='NEMO', exp='test-eORCA1', source="short-3d")
data = reader.retrieve()
print(data.so)
```

### Error messages

```
2026-09-02 20:16:11 :: Regridder :: WARNING  -> Horizontal dimensions not found in the grid path. Please provide a dataset. 
2026-09-02 20:16:12 :: BackendIntakeXarray :: ERROR    -> Dataset is not a dask-backed array. Please verify the data access.

<xarray.DataArray 'so' (depth: 8, ncells: 120184)> Size: 8MB
array([[nan, nan, nan, ..., nan, nan, nan],
       [nan, nan, nan, ..., nan, nan, nan],
       [nan, nan, nan, ..., nan, nan, nan],
```

None

mcadau:
## PR description:
I decided to start again from scratch, to minimize the impacts on the other drivers and `aqua-backend` components. 

Thanks a lot @jhardenberg and @mnurisso for your comments and reviews. 
This follows now, with the same scope as #2935 , the structure defined below: 

- porting of the intake-xarray code we were interested to take into the `intake_drivers/xarray` folder (into `base.py`, `netcdf.py` and `xzarr.py` files)
- no custom openers and datatypes here: intake2 already includes all the structure for netCDF and Zarr (Jost was right in his comment)
- no other drivers changes
- only minimal changes to BackendIntakeXarray just to solve some hacks we were forced to use with `intake-xarray`
- intake-xarray stub in `compat.py` should still be necessary for now, as catalog entries still cite intake-xarray. As such, `compat.py` can be removed once this is fixed in the catalog

close #2958
close #3064

----

 - [x] Changelog is updated.


aqua/core/intake_drivers/xarray/compat.py
```diff
@@ -0,0 +1,42 @@
+"""Compatibility stub for catalogs still referencing the intake_xarray package."""
+
```
oloapinivad, line 2: what's this file? it does not seem like the other standard approach or am I missing something?

mcadau: This is necessary since `Climate-DT-catalog` often cites

```
plugins:
  source:
    - module: intake_xarray
```

in `ci` and `obs` catalog. Once these entries will be removed from the catalog, `compat.py` can be removed

aqua/core/intake_drivers/xarray/netcdf.py
```diff
@@ -0,0 +1,46 @@
...
+        metadata (dict, optional): Catalog metadata for this source.
+        path_as_pattern (bool | str, optional): Treat the path as a ``{field}`` pattern. Defaults to True.
+        storage_options (dict, optional): Parameters passed to the backend file-system (e.g. s3).
+        kwargs: Further parameters forwarded to the reader (e.g. chunks, combine).
+    """
+
+    name = "netcdf"
+
+    def __init__(self, urlpath, xarray_kwargs=None, metadata=None, path_as_pattern=True, storage_options=None, **kwargs):
+        self.xarray_kwargs = xarray_kwargs = fold_use_cftime(dict(xarray_kwargs or {}))
```
jhardenberg: since this is common to the xzarr case, should it not be in IntakeXarraySourceAdapter ?

aqua/core/intake_drivers/xarray/netcdf.py
```diff
@@ -0,0 +1,46 @@
...
+        xarray_kwargs.setdefault("engine", "netcdf4")
```
jhardenberg, line 38: xarray_kwargs is exposed as    self.xarray_kwargs but then it is modified?

aqua/core/intake_drivers/xarray/netcdf.py
```diff
@@ -0,0 +1,46 @@
...
+
+from .base import IntakeXarraySourceAdapter
+from .readers import NetCDFZarrDatasetReader, fold_use_cftime
+
+
+class IntakeNetCDFSource(IntakeXarraySourceAdapter):
+    """Open one or more NetCDF files with xarray, registered as the ``netcdf`` driver.
+
+    Port of ``intake_xarray.netcdf.NetCDFSource`` (intake-xarray 2.0.0). AQUA deltas:
+    ``self.data``, ``self.metadata`` and ``self.xarray_kwargs`` are exposed for the
```
jhardenberg: where is self.metadata defined? I do not see it

aqua/core/intake_drivers/xarray/netcdf.py
```diff
@@ -0,0 +1,46 @@
...
+        storage_options (dict, optional): Parameters passed to the backend file-system (e.g. s3).
+        kwargs: Further parameters forwarded to the reader (e.g. chunks, combine).
+    """
+
+    name = "netcdf"
+
+    def __init__(self, urlpath, xarray_kwargs=None, metadata=None, path_as_pattern=True, storage_options=None, **kwargs):
+        self.xarray_kwargs = xarray_kwargs = fold_use_cftime(dict(xarray_kwargs or {}))
+        xarray_kwargs.setdefault("engine", "netcdf4")
+        data = readers.datatypes.NetCDF3(urlpath, storage_options=storage_options, metadata=metadata)
```
jhardenberg, line 39: I am not sure that this is a real problem, but should we be worried that here it says NEtcdf3 for the datatype?

aqua/core/intake_drivers/xarray/netcdf.py
```diff
@@ -0,0 +1,46 @@
...
+        self.data = data
```
jhardenberg: I would move this up, close to the exposure of self.xarray_kwargs

jhardenberg: by the way, since it is common to the xzarr case, should it nit be in IntakeXarraySourceAdapter ?

aqua/core/intake_drivers/xarray/netcdf.py
```diff
@@ -0,0 +1,46 @@
...
+        if (path_as_pattern is True and "{" in urlpath) or isinstance(path_as_pattern, str):
```
jhardenberg, line 41: AFAIK open_dataset would open a file without dask ... did we test what happens if only one single netcdf file is opened? If this is true then we should make sure that open_fdataset is used always

CHANGELOG.md
```diff
@@ -7,7 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 Unreleased in the current development version (target v1.0.0):
 
 Complete list:
-
+- Remove intake-xarray dependency (#2958)
+- z3fdb intake driver (#2940)
```
jhardenberg: This is already described in the CHANGELOG of aqua-backend branch, did you miss a merge?

CHANGELOG.md
```diff
@@ -7,7 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
...
+- Update submit_aqua_web tool to support kind and separate templates (#2921))
```
jhardenberg: This is described in the CHANGELOG of aqua-backend branch, did you miss a merge?

aqua/core/intake_drivers/xarray/compat.py
```diff
@@ -0,0 +1,42 @@
...
+import importlib.util
+import sys
+import types
+
+from .netcdf import IntakeNetCDFSource
+from .xzarr import IntakeZarrSource
+
+
+# TODO: remove once the catalogs drops the plugin block
+def install_intake_xarray_stub():
```
jhardenberg, line 12: Finje, but where is install_intake_xarray_stub() actually called???

aqua/core/intake_drivers/xarray/netcdf.py
jhardenberg: if this goes to base.py then probably you do not need a separate function

mcadau: You're right: I moved it to the base.py init

aqua/core/intake_drivers/xarray/netcdf.py
mcadau: Corrected

aqua/core/intake_drivers/xarray/netcdf.py
mcadau: This is intake's fault, not ours: however, since we pin netcdf4 as engine so this won't be a problem. I added a comment in any case

aqua/core/intake_drivers/xarray/netcdf.py
mcadau: is defined on the super init (added a comment btw)

aqua/core/intake_drivers/xarray/netcdf.py
mcadau: Yep, done, as in the comment above

aqua/core/intake_drivers/xarray/netcdf.py
mcadau: Tested: identical to `aqua-backend` (only "single file + no chunks anywhere" is non-dask; `to_dask()` always injects `chunks={}`). Forcing `open_mfdataset` would break the eager `read()` (it returns dask even with `chunks=None`)

aqua/core/intake_drivers/xarray/compat.py
mcadau: in the intake_drivers/__init__.py, so the stub is in place before any catalog yaml is parsed

aqua/core/intake_drivers/xarray/netcdf.py
mcadau: Yep, in fact this has been moved to base.py (see comment above)

aqua/core/backend/backend_intake_xarray.py
tovaz92, line 202: is this still valid?

aqua/core/intake_drivers/xarray/readers.py
```diff
@@ -0,0 +1,34 @@
...
+    (e.g. ``combine: by_coords``) while their url may resolve to a single file
+    (glob expansion, date filtering): the stock reader routes such reads to
+    ``xr.open_dataset``, which rejects those kwargs, so they are dropped here.
+    """
+
+    def _read(self, data, **kw):
+        url = data.url
+        single = (isinstance(url, str) and "*" not in url) or (isinstance(url, (list, tuple, set)) and len(url) == 1)
+        if single:
+            kw = {k: v for k, v in kw.items() if k not in MFDATASET_ONLY_KWARGS}
```
tovaz92, line 33: would we in this case loose also potentially useful kwargs such as "preprocess" ? If this is not ok then there should be a raise maybe

tovaz92: On this topic, is it possible to use the Intake `XArrayDatasetReader` (and `XArrayPatternReader`)? Would this simplify things and implementation?

aqua/core/backend/backend_intake_xarray.py
tovaz92, line 24: Maybe the `_get_source_urls` and `_get_xarray_kwargs_from_catalog` can be removed from `CatalogMixin`?

aqua/core/intake_drivers/xarray/base.py
```diff
@@ -0,0 +1,43 @@
...
+
+from .readers import NetCDFZarrDatasetReader
+
+
+class IntakeXarraySourceAdapter(base.DataSource):
+    container = "xarray"
+    name = "xarray"
+    version = ""
+
+    def __init__(self, data, xarray_kwargs=None, metadata=None, reader_class=NetCDFZarrDatasetReader, **kwargs):
```
tovaz92, line 11: Can we use `reader_class=readers.XArrayDatasetReader`? this may avoid the import and simplify to track changes (unless we want to) in `open_dataset` and `open_mfdataset`.

aqua/core/backend/backend_intake_xarray.py
mcadau: Yes it is, and it was already present on aqua-backend before this PR

aqua/core/intake_drivers/xarray/readers.py
mcadau: discarded kwargs always check how to combine multiple datasets, but on a single file there's nothing to combine, so this should be safe. In any case I improved the comment to better explain why the drop is not risky. 

To comment 2: we already use Intake XArrayDatasetReader  (look at the arguments of the class)

aqua/core/intake_drivers/xarray/base.py
mcadau: Same as your second comment of comment no.2 of the review

aqua/core/backend/backend_intake_xarray.py
mcadau: Yep this is correct. You're right. Removed

mnurisso:
Please rebase

aqua/core/intake_drivers/xarray/netcdf.py
```diff
@@ -0,0 +1,47 @@
+"""AQUA-provided intake source for NetCDF data, port of ``intake_xarray.netcdf``."""
+
+from intake import readers
+
+from .base import IntakeXarraySourceAdapter
+
+
+class IntakeNetCDFSource(IntakeXarraySourceAdapter):
+    """Open one or more NetCDF files with xarray, registered as the ``netcdf`` driver.
```
jhardenberg, line 9: I would clone a great part of the intake-xarray docstring which seems much more clear to me:

    urlpath : str, List[str]
        Path to source file. May include glob "*" characters, format
        pattern strings, or list.
        Some examples:
            - ``{{ CATALOG_DIR }}/data/air.nc``
            - ``{{ CATALOG_DIR }}/data/*.nc``
            - ``{{ CATALOG_DIR }}/data/air_{year}.nc``
    chunks : int or dict, optional
        Chunks is used to load the new dataset into dask
        arrays. ``chunks={}`` loads the dataset with dask using a single
        chunk for all arrays.
    combine : ({'by_coords', 'nested'}, optional)
        Which function is used to concatenate all the files when urlpath
        has a wildcard. It is recommended to set this argument in all
        your catalogs because the default has changed and is going to change.
        It was "nested", and is now the default of xarray.open_mfdataset
        which is "auto_combine", and is planed to change from "auto" to
        "by_corrds" in a near future.
    concat_dim : str, optional
        Name of dimension along which to concatenate the files. Can
        be new or pre-existing if combine is "nested". Must be None or new if
        combine is "by_coords".
    path_as_pattern : bool or str, optional
        Whether to treat the path as a pattern (ie. ``data_{field}.nc``)
        and create new coodinates in the output corresponding to pattern
        fields. If str, is treated as pattern to match on. Default is True.
    xarray_kwargs: dict
        Additional xarray kwargs for xr.open_dataset() or xr.open_mfdataset().
    storage_options: dict
        If using a remote fs (whether caching locally or not), these are
        the kwargs to pass to that FS.

aqua/core/intake_drivers/xarray/netcdf.py
```diff
@@ -0,0 +1,47 @@
...
+        xarray_kwargs = dict(xarray_kwargs or {})
+        # intake infers engine="scipy" from the NetCDF3 datatype, and scipy fails on any
+        # other flavour: netcdf4 reads them all (the backend used to force it downstream)
+        xarray_kwargs.setdefault("engine", "netcdf4")
+        data = readers.datatypes.NetCDF3(urlpath, storage_options=storage_options, metadata=metadata)
+
+        if (path_as_pattern is True and "{" in urlpath) or isinstance(path_as_pattern, str):
+            # ``{field}`` patterns become output coordinates: intake has its own reader for that
+            super().__init__(
+                data, xarray_kwargs, metadata, reader_class=readers.XArrayPatternReader, pattern=path_as_pattern, **kwargs
```
jhardenberg, line 44: I really do not like the idea of passing a reader_class to the base reader class. I would rather drop that argument, not initialize self.reader in the super and rather do that here, no?
And there could a common final call to `super().__init__(data, xarray_kwargs, metadata, **kwargs)`

jhardenberg: The chunks fix should then also be here

aqua/core/intake_drivers/xarray/base.py
```diff
@@ -0,0 +1,47 @@
...
+        # ``xr.open_dataset``, which is eager unless ``chunks`` is given
+        if "chunks" not in xarray_kwargs:
+            kwargs.setdefault("chunks", {})
+        self.xarray_kwargs = xarray_kwargs
+        self.data = data
+        self.reader = reader_class(data, **xarray_kwargs, metadata=metadata, **kwargs)
+        super().__init__(metadata=metadata)
+
+    def to_dask(self):
+        if "chunks" not in self.reader.kwargs:
```
jhardenberg, line 32: is this a repetition of the chunks fix above?

aqua/core/intake_drivers/xarray/netcdf.py
```diff
@@ -0,0 +1,47 @@
...
+            )
+        else:
+            super().__init__(data, xarray_kwargs, metadata, **kwargs)
```
jhardenberg, line 47: base.py has NetCDFZarrDatasetReader as a default if reader_class is not specified...so if what we read is not a multiple pattern then NetCDFZarrDatasetReader is used. This one is actually derived from intake's XArrayDatasetReader .... so I find the name a bit confusing. Why not call it XArrayDatasetReaderRobust or similar since it is only a modification of the latter?

aqua/core/intake_drivers/xarray/readers.py
```diff
@@ -0,0 +1,39 @@
...
+    "parallel",
+    "join",
+    "compat",
+    "coords",
+    "data_vars",
+    "combine_attrs",
+)
+
+
+class NetCDFZarrDatasetReader(XArrayDatasetReader):
```
jhardenberg, line 23: See above, I would rename this one

jhardenberg:
@mcadau what is the status of this one?

aqua/core/intake_drivers/xarray/readers.py
mcadau: I thought to rename it to `TolerantXArrayDatasetReader`, as the docstring was already suggesting

mcadau:
Ciao @jhardenberg sorry for the late. I finally worked at your comments. Do you think this is ready to be merged now?

aqua/core/intake_drivers/xarray/netcdf.py
```diff
@@ -0,0 +1,67 @@
...
+        xarray_kwargs.setdefault("engine", "netcdf4")
+        # AQUA always wants lazy data, but intake routes a single file to ``xr.open_dataset``,
+        # which is eager unless ``chunks`` is given
+        if "chunks" not in xarray_kwargs:
+            kwargs.setdefault("chunks", {})
+
+        data = readers.datatypes.NetCDF3(urlpath, storage_options=storage_options, metadata=metadata)
+        if (path_as_pattern is True and "{" in urlpath) or isinstance(path_as_pattern, str):
+            # ``{field}`` patterns become output coordinates: intake has its own reader for that
+            reader = readers.XArrayPatternReader(data, **xarray_kwargs, metadata=metadata, pattern=path_as_pattern, **kwargs)
```
jhardenberg, line 63: Just one question: did you try if the PatternReader could not be usaed in both cases (i.e. always use `xr.open_mfdataset()` ) ? What would happen if we pass an urlpath which has no pattern also to the PatternReader?

jhardenberg:
For the rest, apart from my question above, this looks pretty clean to me.

jhardenberg (approved):

jhardenberg:
@mcadau let's wait with merging this...I am still checking one thing

jhardenberg (commented):
not approving yet, investigating

aqua/core/intake_drivers/xarray/netcdf.py
jhardenberg: Ok, so I tried... the PatternReader actually never seems to work. I actually wonder if we want this? We never use that option in our catalogues and (I believe) we have no test for it

jhardenberg:
Ok, so I believe that in our catalogues (and our tests?) we never use the pattern capability (url like `filename_{year}.nc` or similar). So that call to `XarrayPatternReader` is never triggered. So question is: does it work? Even if we do not use it now it could become handy in the future but looking at the code of intake I do not understand why it should work: if we pass the argument `pattern=True` (and the reader always does that) I see no treatment for the case of `pattern` being a boolean in the intake sources... it seems always to expect a string. See https://github.com/intake/intake/blob/master/intake/readers/readers.py around line 1484

If this is confirmed that would be an original intake-xarray bug (since those lines are copied from intake-xarray), not ours, but then should we keep those lines if in the end it does not work?
We should check this.

oloapinivad:
what this PatternReader is about? It allows you to filter by a certain variable the file input? We have implemented the `filter_key` option to do something back in the days, it would be cool if this is something already available in intake-xarray

jhardenberg:
> what this PatternReader is about? It allows you to filter by a certain variable the file input? We have implemented the `filter_key` option to do something back in the days, it would be cool if this is something already available in intake-xarray

What it is supposed to do is to allow writing as a source:

`urlpath: era5_{variable}_{time}.nc` 
and this would then read a bunch of files like `era5_2t_.nc, era5_2t_19800101.nc, era5_2t_19800102.nc` etc ... even if they did not contain the time axis it would recreate it .... 

It is something which we definitively are not using (and I suspect it does not work anyway).
So I discussed this with @mcadau . Our plan is: he will do a brief series of tests to understand if this feature works at all. If (as I expect) it does not, we will comment those lines and leave it out of this PR. We can then open another PR to make it work (and write corresponding tests) since it could maybe be useful to read certain input files from obs without having to preprocess them.

jhardenberg:
Here more examples of what this hsould in theory be able to do (here it refers to csv but it is applied also to nc)


    >>> reverse_format('data_{year}_{month}_{day}.csv', 'data_2014_01_03.csv')
    {'year': '2014', 'month': '01', 'day': '03'}
    >>> reverse_format('data_{year:d}_{month:d}_{day:d}.csv', 'data_2014_01_03.csv')
    {'year': 2014, 'month': 1, 'day': 3}
    >>> reverse_format('data_{date:%Y_%m_%d}.csv', 'data_2016_10_01.csv')
    {'date': datetime.datetime(2016, 10, 1, 0, 0)}
    >>> reverse_format('{state:2}{zip:5}', 'PA19104')
    {'state': 'PA', 'zip': '19104'}

So if we write

`urlpath: data_{time:%Y_%m_%d}.nc` this will read file `data_1990_01_01.nc, data_1990_01_02.nc` even if they contain no time axis .... in theory powerful stuff

mcadau:
Ok, i performed many tests, both on this branch and in the current `main`. @jhardenberg your suspicion is right: `pattern=True` is passed as a boolean, while intake's `XArrayPatternReader` needs a string when `data.url` is a list. That comes from `intake-xarray` and we copied it. But this is not the only problem, and it's not the first one you hit through the Reader: 
1) on our backend, `_check_netcdf_files_exist` uses `glob`, which doesn't understand `{field}` (it knows *, ? and [...]). This leads to `NoDataError` at `Reader` level; 
2) from `intake-xarray`, `pattern=True` passed as a boolean leads to `TypeError: expected str, got bool` as soon as `data.url` is a list, because our backend always makes it so; 
3) on intake, more than one field (like `era5_{variable}_{time}.nc`) leads to `ValueError` on `concat_dims`; 
4) on intake again, a pattern matching a single file leads to `TypeError`, because `open_dataset` receives `concat_dim`.

`main` and this branch act the same in every case so there is no regression, and through the AQUA Reader `urlpath: data_{time:%Y%m%d}.nc` leads to `NoDataError`, while `urlpath: data_*.nc` + `path_as_pattern: data_{time:%Y%m%d}.nc` is ok and time axis is rebuilt from the file names, so your example can already be read with the `Reader`, as long as the glob goes in `urlpath` and pattern in `path_as_pattern`.

That's why I wouldn't comment out those lines: first, none of our catalogs uses patterns, so keeping them keeps no risk;
secondly, commenting them out fixes none of the failing cases, and breaks the ones that work today, including the `glob` + `path_as_pattern` form above. `path_as_pattern` would be silently ignored, and xarray would fail with `Could not find any dimension coordinates to use to order the Dataset objects for concatenation`.

My opinion is that we can merge this PR as it is and open a new issue/follow-up PR afterwards.

jhardenberg:
thanks @mcadau for the analysis. 

I second your suggestion: Let's merge this PR (which is no worse or better than intake-xarray for the PatternReader). I would immediately open a (low priority) issue to remember that if we would like that feature to really work some more work is needed.

oloapinivad:
Please remember for the future that to fully test the code you need to recreate the conda lock. otherwise the environment is the same as before.

mcadau:
Yep, sorry! Is it already in the CI steps?

PR #2960, opened 2026-07-15
[Fix] Correct initialization of area selection in fldstat

mnurisso:
## PR description:

I noticed that with areas=False the area selection does not work. This because its constructor happens after the check on the presence of an area in Fldstat, but this is not necessary for the area selection. Therefore I'm moving above the constructor so that it's built also in this specific setup (which unluckily is mine for TRMM)

 - [x] Changelog is updated.


mnurisso:
Added a test and the changelog, PR is ready

silviacaprioli (approved):

PR #2968, opened 2026-07-24
[Backend] Aqua backend `ConfigPath()` refactor

oloapinivad:
## PR description:

This is a tentative reworking of the `ConfigPath` so that it splits strictly intake operations from more generic data operations. For now is fully backward compatible.


mnurisso:
I think it has to be rebased to the `aqua-backend` branch

mnurisso:
Everything seems ok, so far this is only code cleaning but there is no computational advantage, the amount of code executed is the same as before. Do we want to split the call in a follow up PR?

oloapinivad:
This is an initial split but I want to keep working on this in next days (I am already working on it). I want to use `ConfigContext` everywhere and call `ConfigCatalog` only when needed.

oloapinivad:
Ok, the idea is in place, we probably need to refine and clean up quite a lot of rubbish, but the draft is there. Waiting for comments.

aqua/core/configurer/context.py
```diff
@@ -0,0 +1,127 @@
...
+            locator = ConfigLocator(filename=filename, configdir=configdir, logger=self.logger)
+        if packages is None:
+            packages = ConfigPackages(logger=self.logger)
+        self.locator = locator
+        self.packages = packages
+        self.configdir = self.locator.configdir
+        self.config_file = self.locator.config_file
+        self.logger.debug("Configuration file found in %s", self.config_file)
+        self.config_dict = load_yaml(self.config_file)
+        self.machine = None
```
tovaz92: Should this be updated whe calling `get_machine(self)`? Like `self.machine = self.get_machine()`
I would add a small test on this for safety.

aqua/core/configurer/catalog.py
```diff
@@ -0,0 +1,407 @@
...
+            intake_vars (dict): the intake catalog variables
+        """
+        # loading the grid defintion file
+        machine_file = load_yaml(self.machine_file)
+        machine_paths = {}
+
+        # get information on paths
+        if self.paths.machine in machine_file:
+            machine_paths = machine_file[self.paths.machine]
+        else:
```
tovaz92: this can be an `elif`

aqua/core/configurer/configpath.py
```diff
@@ -1,484 +1,116 @@
...
-            self.logger.debug("Catalog found in %s file are %s", self.config_file, base["catalog"])
-            return base["catalog"]
+    @property
+    def catalog_available(self):
+        return self._catalog.catalog_available
 
-        raise FileNotFoundError(f"Cannot find the basic configuration file {self.config_file}!")
+    @property
+    def base_available(self):
+        return self._catalog.base_available
```
tovaz92: Not sure if this method does exist

aqua/core/configurer/configpath.py
```diff
@@ -1,484 +1,116 @@
...
-
-    #     # Search for the dictionary key in the key_string
-    #     for key, value in platform_dict.items():
-    #         if key in platform_name:
-    #             self.logger.debug('%s machine identified!', value)
-    #             return value
-
-    #     self.logger.debug('No machine identified, still unknown and set to None!')
-    #     return None
+        """See `ConfigCatalog.get_base`."""
```
tovaz92: Same for this, not sure if it exist

aqua/core/reader/catalog.py
```diff
@@ -37,7 +34,7 @@ def show_catalog_content(
         dict: Dictionary with catalog names as keys and nested dict structure
               as values.
     """
-    config = ConfigPath(configdir=configdir, catalog=catalog_name, loglevel=loglevel)
+    config = ConfigCatalog(catalog=catalog_name, loglevel=loglevel)
```
tovaz92, line 37: This might need
```
context = ConfigContext(configdir=configdir)
config = ConfigCatalog(configcontext=context,..)
```
to account for custom configuration folder

oloapinivad: the problem is that configdir is not exposed in show catalog content, shall we expose it?

aqua/core/configurer/configpath.py
oloapinivad: correct, these are old methods. fixed

aqua/core/configurer/configpath.py
oloapinivad: correct, these are old methods. fixed

aqua/core/configurer/catalog.py
oloapinivad: fixed

aqua/core/configurer/context.py
oloapinivad: I refactored this, now it should be a cached property

aqua/core/reader/catalog.py
tovaz92: Most of the time the method will be called without arguments, so we can also leave as it is now

aqua/core/configurer/context.py
```diff
@@ -0,0 +1,131 @@
...
+            loglevel (str): The logging level. Defaults to 'warning'.
+            locator (ConfigLocator | None): An optional ConfigLocator instance.
+        """
+
+        # set up logger
+        self.logger = log_configure(log_level=loglevel, log_name="ConfigContext")
+
+        # get the configuration directory and its file
+        self.locator = locator or ConfigLocator(filename=filename, configdir=configdir, logger=self.logger)
+        self.packages = packages or ConfigPackages(logger=self.logger)
```
tovaz92, line 51: Is this going to be used in the future? I am not sure how it is used at the moment

tovaz92 (approved):

silviacaprioli:
merging

PR #2972, opened 2026-07-27
[Datamodel] Improve coordinate range operations

oloapinivad:
## PR description:

This extends the #2957 , still working on it. 

Close #issue_number


oloapinivad:
So I made a few updates, but I am still confused by tests which seems to be written without actually really exploiting the philosophy of CoordTransformer. I will double check this in next days. @maqsoodrajput feel free to comment!

oloapinivad:
So I should have been able to make it works. I removed some tests and rerranged some others since it seemed a bit ill-defined, and introduce an internal naming scheme which is useful to avoid mistakes on names. 
I simplified a bit a couple of passages, and it looks efficient and easier to follow now. @maqsoodrajput let me know what do you think, make some comments and potentially approve it so that we can merge it in your and then merge into main.
Only open question, what is the rationale for the tolerance threshold to 1.5e-5?

maqsoodrajput (approved):
Looks good to me!

maqsoodrajput:
> So I should have been able to make it works. I removed some tests and rerranged some others since it seemed a bit ill-defined, and introduce an internal naming scheme which is useful to avoid mistakes on names. I simplified a bit a couple of passages, and it looks efficient and easier to follow now. @maqsoodrajput let me know what do you think, make some comments and potentially approve it so that we can merge it in your and then merge into main. Only open question, what is the rationale for the tolerance threshold to 1.5e-5?

maqsoodrajput:
> > So I should have been able to make it works. I removed some tests and rerranged some others since it seemed a bit ill-defined, and introduce an internal naming scheme which is useful to avoid mistakes on names. I simplified a bit a couple of passages, and it looks efficient and easier to follow now. @maqsoodrajput let me know what do you think, make some comments and potentially approve it so that we can merge it in your and then merge into main. Only open question, what is the rationale for the tolerance threshold to 1.5e-5?

regarding the tolerance threshold of 1.5e-5, it's basically `0.00001 degrees (1e-5) is approx 1.1 meters`. So if we go to very small grid spacing the code won't break. Just as a safeguard.

PR #2926, opened 2026-06-29
[Backend] The big backend refactor main branch

mcadau:
## PR description:
Work in progress.

----

 - [ ] Tests are included if a new feature is included.
 - [ ] Documentation is included if a new feature is included.
 - [ ] Docstrings are updated if needed.
 - [ ] Changelog is updated.
 - [ ] Notebooks which requires changes are updated.
 - [ ] Run Cross check tests for AQUA-diagnostics [AQUA-diagnostics cross-check workflow](https://github.com/DestinE-Climate-DT/AQUA-diagnostics/actions/workflows/aqua-diagnostics-cross-check.yml)
 - [ ] environment.yml and pyproject.toml are updated if needed, together with the lumi installation tool. Installation instructions are updated if new conda dependencies are added.
 - [ ] Check `ruff` and `pre-commit` commands: `ruff check . --no-cache`, `pre-commit run --all-files` and `ruff format --check . --no-cache` are successful.

PR #2981, opened 2026-08-04
[Backend] Regridder metadata

oloapinivad:
## PR description:

This PR targets an old problem of propagating the reader metadata to the filenames for a proper weights and area creation. It introduces a small class to rationalise the problem and the creates a series of fallback that should allow a proper usage of the `Regridder` class alone or with the `path` argument.


oloapinivad:
With the last modification, even with a hack, we can use the xarray path based access with the regridder 

```python
from aqua import Reader
reader = Reader(path="/Users/paolo/Desktop/aqua-dvc/data-tests/models/IFS/long/regridded_r18x9.nc", regrid='r100')
data = reader.retrieve()
reader.regrid(data)['2t'].isel(time=0).plot()
```

aqua/core/backend/backend_factory.py
```diff
@@ -159,6 +159,9 @@ def _select_backend_xarray(self):
         Activate the driver for xarray
         """
         self.driver = "xarray"
+        self.logger.warning("Using default path in local folder for areas, weights, and grids.")
```
mnurisso, line 164: Add a comment before merging, we could have a path set in the config-aqua.yaml which in this case should be used and should be updated in this piece of code.

aqua/core/regridder/filename_handler.py
```diff
@@ -0,0 +1,136 @@
...
+        return f"weights_unknown_{method}_{targetgrid}_l{level}.nc"
+
+    def _insert_metadata_params(self, filename, regridder_metadata):
+        """Insert extra params from DEFAULT_WEIGHTS_AREAS_PARAMETERS into filename."""
+        if not regridder_metadata:
+            return filename
+        for param in DEFAULT_WEIGHTS_AREAS_PARAMETERS:
+            value = getattr(regridder_metadata, param, None)
+            if value is not None:
+                filename = re.sub(r"\.nc", f"_{param}{value}.nc", filename)
```
mnurisso: substitute with filename = re.sub(r"\.nc$", f"_{param}{value}.nc", filename) to ensure no random .nc is substituted

oloapinivad:
@mnurisso tests are failing here for some weird reason after your last updated. do we have any idea what this is not working anymore?

mnurisso:
> @mnurisso tests are failing here for some weird reason after your last updated. do we have any idea what this is not working anymore?

Test should be fine now after we fixed the tests in the #2926 
I suggest to align this PR to that branch and rerun tests.

PR #3005, opened 2026-08-18
Enhance to_list function to handle ruamel.yaml CommentedSeq input

mnurisso:
## PR description:

When reading a list of levels from a Yaml file python reads a `CommentedSeq` object. While in most of the situation this is fine, it creates an error when this list is then used to define the levels to be inserted in the request for the GSV intake driver.
The PR adds in the `to_list` function the possibility to convert to a python list also the `CommentedSeq` so that no other parts or future parts of code need to be modified.

 - [x] Changelog is updated.


aqua/core/drop/output_path_builder.py
```diff
@@ -108,7 +108,9 @@ def build_filename(self, var=None, level=None, year=None, month=None, day=None,
         year = "*" if year is None else year
 
         # specific case for a specific list of levels
-        level = to_list(str(level)) if level is not None else None
+        level = to_list(level) if level is not None else None
+        if level:
+            level = [str(l) for l in level]  # Convert all levels to strings
```
silviacaprioli, line 113: Is this necessary?

jhardenberg (approved):
looks good to me

aqua/core/drop/output_path_builder.py
mnurisso: we want to be sure that int and float are always correctly converted to string before being inserted in the filename

silviacaprioli (approved):

PR #3006, opened 2026-08-18
Add memory option to submit-aqua-web

jhardenberg:
## PR description:

The default job script used submit-aqua-web did not specify memory, resulting in only 120GiB allocated.
This now allows to specify how much memory on the node we wish to use
  
@CostanzaD @mcadau 

----

 - [x] Changelog is updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Unreleased in the current development version (target v1.0.0):
 ClimateDT workflow modifications:
 
 Complete list:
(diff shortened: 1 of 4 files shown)


jhardenberg:
The fact that by default we were using only 120GB (but requesting for dask 32 workers with 7GB each) also could explain the problem with workers dying which we observed in tehse days. Two runs using --mem=224G have both successfully run with no issues on lumi

mcadau (approved):

PR #3020, opened 2026-08-21
Intake_gsv treatment of fdb_info_file

### LINKED ISSUE #3019: fdb_info_file is not automatically read
### machine

MN5

### catalog

Others (specify)

### version

main

### What happened?

During workflow integration testing:
https://gitlab.earth.bsc.es/digital-twins/de_340-3/workflow/-/merge_requests/1394#note_476918

we noticed that, when auto is set in the catalog entry, the LRA JOB does not seem to automatically read the fdb_info_file containing the experiment's start and end dates.

@jhardenberg,

```python

```

### Error messages

_No response_

None

jhardenberg:
## PR description:

This introduces a test which would fail with the current main and a small fix which should address issue #3019 

Close #3019

---

 - [x] Changelog is updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Unreleased in the current development version (target v1.0.0):
 ClimateDT workflow modifications:
 
 Complete list:
+- Fix intake_gsv fdb_info_file treatment (#3020)
 - Fix and rewrite get_eccodes_attr for paramId case (#3008)
 - Add memory option to submit-aqua-web (#3006)
 - Synchronize the available statistics in DROP with the TimStat class (#2991)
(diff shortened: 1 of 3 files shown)


jhardenberg:
It is actually not entirely clear to me why we had that additional check on self.dummy_run. It is not clear to me if there are further implications. I could imagine that there may be problems on machines where we have a fdb_info_file defined in the catalog but we wish to use polytope: this should never happen though, since our general catalogs do not have the fdb_info_file defined, it is a thing used only by the workflow

silviacaprioli (approved):

jhardenberg:
@silviacaprioli ok all tests have now worked. I will merge. Let's discuss a new release for Damien on Monday then

PR #3031, opened 2026-08-26
DVC new configuration

oloapinivad:
## PR description:

This tries to implement the new ECMWF DVC structure, which requires a new key and new new user and password.
I manually updated the secretes and added a new `.pem` file that has to be configured. This are required for every user, seriously limiting the capability that DVC will have in distributing the data.


oloapinivad:
This is not simple, but with this it seems that now we have back again working DVC

oloapinivad:
@silviacaprioli @mnurisso this brings back DVC, I am porting this to diagnostics too

silviacaprioli (approved):

PR #3038, opened 2026-08-28
Coordinate data model scanner

### LINKED ISSUE #2581: Vertical coordinate scanner utility through `Data Model`
Utility function, possibly also inside the Reader, to scan the name of a coordinate in the chosen data model, so that diagnostics can use this utility instead of guessing an hard-coded coordinate name.

The datamodel is already computing the vertical coordinate but we also have a `find_vert_coord()` function to do the same. So this function is a duplicate of functionality already existing and it would be better to ask the datamodel directly

mnurisso:
## PR description:

As requested in #2581 I've added a function that allow to explore the data model yaml file and get the name field, so that if for a data model the ocean depth has to be level the function will return level, allowing the diagnostics to use this function to be able to not hardcode any coordinate name but to be datamodel aware.

The function supports multiple data models and the possibility to set a default if the datamodel fails (e.g. the coordinate is not defined there)

Close #2581 

 - [x] Changelog is updated.

### DIFF
--- a/aqua/core/data_model/coord_utils.py
+++ b/aqua/core/data_model/coord_utils.py
@@ -91,6 +91,26 @@ def get_data_model(name: str = "aqua"):
     return _load_data_model(name)
 
 
+def scan_coord(internal_name: str, default: str = None, model: str = "aqua") -> str:
+    """
+    Get the coordinate name used by a given data model for an internal coordinate type
+    (e.g. 'depth', 'isobaric'), so that diagnostics do not need to hard-code coordinate names.
+
+    Args:
+        internal_name (str): Internal coordinate type (e.g. 'depth', 'isobaric', 'latitude').
+        default (str): Fallback name returned if the coordinate is not defined in the data model.
+        model (str): Data model name. Default is 'aqua'.
+
+    Returns:
+        str: Coordinate name as defined by the data model, or the default if not found.
+    """
+    try:
+        coord = get_data_model(model)["data_model"][internal_name]
+    except (FileNotFoundError, KeyError):
+        return default
+    return coord.get("name", default)
+
+
 # Function to get the conversion factor
 def units_conversion_factor(from_unit_str, to_unit_str):
     """
--- a/aqua/core/data_model/coord_utils.py
+++ b/aqua/core/data_model/coord_utils.py
@@ -106,8 +106,10 @@ def scan_coord(internal_name: str, default: str = None, data_model: str = "aqua"
     """
     try:
         coord = get_data_model(data_model)["data_model"][internal_name]
-    except (FileNotFoundError, KeyError):
+    except FileNotFoundError:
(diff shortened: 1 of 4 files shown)


aqua/core/data_model/coord_utils.py
```diff
@@ -91,6 +91,26 @@ def get_data_model(name: str = "aqua"):
...
 
+def scan_coord(internal_name: str, default: str = None, model: str = "aqua") -> str:
+    """
+    Get the coordinate name used by a given data model for an internal coordinate type
+    (e.g. 'depth', 'isobaric'), so that diagnostics do not need to hard-code coordinate names.
+
+    Args:
+        internal_name (str): Internal coordinate type (e.g. 'depth', 'isobaric', 'latitude').
+        default (str): Fallback name returned if the coordinate is not defined in the data model.
+        model (str): Data model name. Default is 'aqua'.
```
oloapinivad: what about `data_model`? `model` has an heavy mening in the  `Reader`

aqua/core/data_model/coord_utils.py
```diff
@@ -91,6 +91,26 @@ def get_data_model(name: str = "aqua"):
     return _load_data_model(name)
 
 
+def scan_coord(internal_name: str, default: str = None, model: str = "aqua") -> str:
+    """
+    Get the coordinate name used by a given data model for an internal coordinate type
+    (e.g. 'depth', 'isobaric'), so that diagnostics do not need to hard-code coordinate names.
```
oloapinivad, line 97: we have the `internal_name` defined, I would add a error trap to ensure that users call the right internal name.

mnurisso: added

aqua/core/data_model/coord_utils.py
mnurisso: renamed

mnurisso:
No gitlab available, but in principle now the PR is ready and waiting for #2957 to be merged before merging this one

mnurisso:
@oloapinivad I remember that we discussed some extra adaptation to be done after the merge of #2957 but honestly it is not clear what to do extra so I'll wait for a hint or a green light for merging the PR.

What is missing here is to utilize the new function in the Reader so that we get rid of hardcoded coordinate usage in fldmean and similar, but this is extra wrt to the purpose of this PR.

mnurisso:
A bit of refinement, all the internal names are now defined as `AQUA_COORDS`, this allow usage in my `scan_coords` and code simplifications in `CoordIdentifier`, so that a new internal name will require no code modification.

I have a doubt, we have now `DEFAULT_COORD_NAMES` and a yaml file in `config/data_model/coords_default.yaml`. Is the second still necessary? Can be deleted?

oloapinivad (approved):

PR #3041, opened 2026-08-29
Solution for trender of irregular monthly data 

### LINKED ISSUE #3040: Trend calculation fails if months are irregularly spaced
### machine

HPC2020

### catalog

Others (specify)

### version

main

### What happened?

If we try to compute trends with `Trender.trend()` of monthly data where the central dates are irregularly spaced (e.g. '1890-01-16 12:00:00', '1890-02-15 00:00:00',    '1890-03-16 12:00:00',] then the function will fail in its use of the Pandas .method `infer_freq`

I have a solution, see attached PR

```python

```

### Error messages

_No response_

None

jhardenberg:
## PR description:

This solves #3040.
It also introduces a test which would fail if we did not have this solution.

Close #3040 

----

 - [x] Changelog is updated.


aqua/core/reader/trender.py
```diff
@@ -99,14 +100,21 @@ def _coeffs(
...
-            inferred_freq = time_values.inferred_freq
+            inferred_freq = xarray_to_pandas_freq(time_values)
             self.logger.debug("Inferred frequency for 'time' dimension: %s", inferred_freq)
             if inferred_freq is None:
                 raise ValueError(
                     "Inferred frequency for 'time' dimension is None. "
                     "Ensure that the time dimension has a pandas compatible frequency."
                 )
-            offset = pd.tseries.frequencies.to_offset(inferred_freq)
+            if inferred_freq in ["YS", "AS", "Y", "A", "YE"]:
```
oloapinivad: I am sorry but this is very ugly: it depends on a set on arbitrary names, and might fail if some combination is not covered. are we sure that there is not better option for this?

jhardenberg: all this is now in `pandas_freq_to_offset()` in `util/time.py`. The frequencies being matched are standard Pandas frequencies for the yearly case (the list could be longer). This is actually enough because it covers the frequencies generated by `xarray_to_pandas_freq()`

jhardenberg: I did not find better solutions. We coudl ask xarray_to_pandas_freq to also return directly an offset, but this would change the API and this function is already being used in AQUA-diagnostics

aqua/core/util/time.py
```diff
@@ -77,35 +76,81 @@ def frequency_string_to_pandas(freq):
...
+
+    if hasattr(time_coord, "to_index"):
+        time_index = time_coord.to_index()
+    elif isinstance(time_coord, pd.Index):
+        time_index = time_coord
     else:
-        return f"{hours}h"
+        time_index = pd.to_datetime(time_coord)
+
+    diffs = pd.Series(time_index).diff().dropna()
```
oloapinivad, line 113: will it work if there is a time dimension which has lenght=1?

aqua/core/util/time.py
```diff
@@ -77,35 +76,81 @@ def frequency_string_to_pandas(freq):
...
+    if isinstance(xdataset, xr.Dataset):
+        if dim not in xdataset.coords:
+            return None
+        time_coord = xdataset[dim]
+    elif isinstance(xdataset, xr.DataArray):
+        if dim in xdataset.coords:
+            time_coord = xdataset[dim]
+        else:
+            time_coord = xdataset
+    else:
```
oloapinivad, line 103: what about `elif isinstance(pd.Index)` and then a raise? or will it raise anyway due to the type formatting?

aqua/core/util/time.py
```diff
@@ -77,35 +76,81 @@ def frequency_string_to_pandas(freq):
...
+
+    diffs = pd.Series(time_index).diff().dropna()
+    if diffs.empty:
+        return None
+
+    delta = diffs.median()
+    if pd.isna(delta):
+        return None
+
+    if not isinstance(delta, pd.Timedelta):
```
oloapinivad, line 121: uhm, the results of `.diff` can be an object which is not `pd.Timedelta`? it always starts from a `pd.Series`

aqua/core/util/time.py
```diff
@@ -77,35 +76,81 @@ def frequency_string_to_pandas(freq):
...
+        delta = pd.Timedelta(delta)
+
+    if delta <= pd.Timedelta(0):
```
oloapinivad, line 124: perhaps a warning, if the delta is negative?

aqua/core/util/time.py
```diff
@@ -77,35 +76,81 @@ def frequency_string_to_pandas(freq):
...
+        return None
+
+    # Calendar-scale intervals
+    if delta >= pd.Timedelta(days=350):
```
oloapinivad, line 128: 360? I don't think it exists calendar with less that 360 days....

aqua/core/util/time.py
```diff
@@ -77,35 +76,81 @@ def frequency_string_to_pandas(freq):
...
+        years = int(round(delta.total_seconds() / (365.25 * 86400)))
```
oloapinivad, line 129: I suspect there is no easy remedy, but this will fail for non-gregorian calendars, correct?

aqua/core/util/time.py
```diff
@@ -77,35 +76,81 @@ def frequency_string_to_pandas(freq):
...
+        return "YS" if years <= 1 else f"{years}YS"
+    if pd.Timedelta(days=25) <= delta <= pd.Timedelta(days=35):
```
oloapinivad, line 131: all those numbers are arbitrary but let's leave with that

aqua/core/util/time.py
```diff
@@ -77,35 +76,81 @@ def frequency_string_to_pandas(freq):
...
+        return "MS"
+    if pd.Timedelta(days=6) <= delta <= pd.Timedelta(days=8):
+        return "W"
+    if delta >= pd.Timedelta(hours=20):
```
oloapinivad, line 135: well, a a day is always defined with 24 hours I hope

oloapinivad (approved):
I made some small comments, the only questionable issue is the problem with non-gregorian calendar which I am not sure it will work

aqua/core/util/time.py
jhardenberg: actually it will work, because all non-gregorian calendars (with 360, 365, 366 days) will round to years=1

aqua/core/util/time.py
jhardenberg: I comment here even if outdated, but the new version has teh same idea. 
Sometimes central dates are centered irregularly... with 350 we are safe (but I agree that it is arbitrary - but it will work)

aqua/core/util/time.py
jhardenberg: the margins are a bit larger than needed but this should not fail and covers cases where the central dates of months are spaced irregularly (e.g.  February 14 - January 16 = 28 days). The margins are larger than needed but why risk it?

aqua/core/util/time.py
jhardenberg: now captured at the beginning, there is a test if length smaller than 2

aqua/core/util/time.py
jhardenberg: I believe that this is what will happen, the type formatting will intercept it

jhardenberg: I checked, actually not, the type formatting is only a hint, python does not enforce it.

jhardenberg: My AI says that this will work in most cases. The function even works if I pass an iterable (e.g. a list) of strings.
It crashes if I pass a strange non-iterable when it tries to do len(time_index) but that I think is what it should do. Probably no need to intercept this further

aqua/core/util/time.py
jhardenberg: outdated

aqua/core/util/time.py
jhardenberg: again, this is a safeguard for cases where two following dates are centered differently. Not sure it covers all cases, but I think we can leave it.

oloapinivad (approved):

mnurisso:
> I made some small comments, the only questionable issue is the problem with non-gregorian calendar which I am not sure it will work

In principle the Reader always convert to Gregorian, so there should be no issue

PR #3043, opened 2026-08-30
Update load container for both core and diagnostics

### LINKED ISSUE #2532: Load container script to be updated

jhardenberg:
## PR description:

This is an alternative to #2873 where `load_aqua_container.sh` has been expanded a bit more and is now able to load both aqua core or aqua diagnostics containers (and I added a '--list functionality).
In this version it is assumed that it stays in core (but it could be moved to diagnostics)

New options are --list / -l and --diagnostics / -d (by default it will look for AQUA core containers).

The paths on the varuios machines which are not lumi  have to be checked / adapted.
The 'native' option at the moment probably works only for core, to be fixed. Work in progress.

Close #2532 
Close #2873

 - [x] Changelog is updated.


cli/aqua-container/load-aqua-container.sh
```diff
@@ -22,10 +22,12 @@ usage() {
                                  Machine supported are lumi, levante and MN5
 
     Options:
-        --native                 Enable native mode: AQUA will read from native env variable.
-        -s <script>              Execute an executable bash or python script.
-        -c <command>             Execute a shell command.
-        -v <version>             Specify the AQUA container version (default: "latest").
+        -n, --native             Enable native mode: AQUA will read from native env variable.
+        -r, --core               Load aqua-core container instead of aqua-diagnostics (default)
```
mnurisso: here is written default aqua-diagnostics, but below container_type is aqua-core

cli/aqua-container/load-aqua-container.sh
```diff
@@ -66,8 +72,12 @@ parse_machine() {
     # Process each option
     while true; do
         case "$1" in
+            -d|--diagnostics)
```
mnurisso, line 79: this option does not match the help description which states the opposite

cli/aqua-container/load-aqua-container.sh
```diff
@@ -94,18 +104,18 @@ parse_machine() {
...
+            echo "ERROR: $aqua_title directory is not set!"
             exit 1
         fi
         # Check if AQUA is set and the file exists
         echo "Selecting native AQUA path: $AQUA"
         echo "Please use this with caution since it is not how the container is meant to be used!"
         echo "Remember to run: pip install -e `$AQUA` once you are in the container"
         if [ ! -d "$AQUA" ]; then
-            echo "ERROR: The AQUA directory does not exist at: $AQUA"
+            echo "ERROR: The $AQUA_title directory does not exist at: $AQUA"
```
mnurisso: is aqua_title case sensitive?

cli/aqua-container/load-aqua-container.sh
```diff
@@ -134,19 +144,19 @@ parse_machine() {
...
-function setup_container_path(){
+function setup_folder(){
     machine=$1
     case "$machine" in
         "lumi")
-            AQUA_folder="/project/project_465000454/containers/aqua"
+            AQUA_folder="/project/project_465002727/containers/$container_dir"
             ;;
 
         "levante")
```
mnurisso: this is not adapted in levante so far. Additionally we have a new project there, what we should do? A new path in the new project?

mnurisso (changes requested):
Some changes or decisions are required before a merge.
Also more in general I see the point of the script, but it should go asap in the aqua-climatedt repo

cli/aqua-container/load-aqua-container.sh
jhardenberg: indeed, the help is wrong, fixed

cli/aqua-container/load-aqua-container.sh
jhardenberg: the help was wrong, this stands

cli/aqua-container/load-aqua-container.sh
jhardenberg: good catch, fixed

cli/aqua-container/load-aqua-container.sh
jhardenberg: Where are containers in levante now? We should adapt the path indeed

jhardenberg:
@mnurisso the 'native' mode part is only written for AQUA-core. I have a hard time to figure out what to do in the AQUA-diagnostics case ... help?

cli/aqua-container/load-aqua-container.sh
mnurisso: I think it's in my work folder of the nextgems project. Additionally there is no usage of aqua containers on levante that I'm aware of, maybe just drop it if not needed?

jhardenberg: I could not find where containers are available on levante, so for now I commented levante. I added in any case a -p option to specify where to find the containers if needed (so the script should work on any machine)

jhardenberg:
The 'native' mode (which is a quite exotic thing anyway can stay but it is mainly  intended for the core only (actually it reads an AQUA env variable, that could also point to AQUA-diagnostics in theory.
@mnurisso opinion welcome

mnurisso:
> The 'native' mode (which is a quite exotic thing anyway can stay but it is mainly intended for the core only (actually it reads an AQUA env variable, that could also point to AQUA-diagnostics in theory. @mnurisso opinion welcome

I've never used it, so I'd simply leave it as it is and not put extra effort on that feature.

mnurisso (approved):

mnurisso:
As mentioned above, I think this can be merged as it is now, without further effort.

PR #3072, opened 2026-09-04
More robust date selection (also beyond 2262)

### LINKED ISSUE #3071: Despite time in us, string date selection beyond 2262 fails
### machine

Others

### catalog

ci

### version

main

### What happened?

When we read datasets time is now represented in microseconds ("us") to allow dates beyond 2262. Despite this using the reader to sepcify startdate or enddates with dates beyond 2262 fails.

The following code

````
from aqua import Reader
reader = Reader(catalog='ci', model='IFS', exp='test-tco79', source="short")
data = reader.retrieve(startdate="2300-01-01", enddate="2300-02-01")
````

produces 

````
OutOfBoundsDatetime: Cannot cast 2300-01-02 00:00:00 to unit='ns' without overflow.
````

The problem lies in the fact that by default xarray converts string dates to a numpy.datetime64 object in nanoseconds. 
 
If we do not use strings but rather explicit microsecond datetime64 to specify startdate and enddate it works as expected

```
data = reader.retrieve(startdate=np.datetime64("2300-01-01", "us"), enddate=np.datetime64("2300-02-01", "us"))
```

In any case we should fix this internally in the backed. First I will write a test to detect this though.

```python

```

### Error messages

_No response_

None

jhardenberg:
## PR description:

This will fix the issue described in #3071 
Work in progress. First of all I add a test which will currently fail.

Close #3071 

----

 - [x] Changelog is updated.

### DIFF
--- a/aqua/core/backend/backend.py
+++ b/aqua/core/backend/backend.py
@@ -1,6 +1,7 @@
 from abc import ABC, abstractmethod
 
 import dask.array as da
+import numpy as np
 import xarray as xr
 from smmregrid import GridInspector
 
@@ -160,7 +161,7 @@ class Backend(ABC):
 
     def _seldate(self, data: xr.Dataset, startdate: str = None, enddate: str = None):
         """Store date bounds for lazy application."""
-        return data.sel(time=slice(startdate, enddate))
+        return data.sel(time=slice(np.datetime64(startdate, "us"), np.datetime64(enddate, "us")))
 
     def _sellevel(
         self,
--- a/aqua/core/backend/backend.py
+++ b/aqua/core/backend/backend.py
@@ -164,7 +164,6 @@ class Backend(ABC):
         t_start = pd.Timestamp(startdate) if startdate else None
         # This guarantees that the end of year, day, hour etc is correctly selected
         t_end = pd.Period(enddate).end_time if enddate else None
-        # t_end = pd.Timestamp(enddate) if enddate else None
         return data.sel(time=slice(t_start, t_end))
 
     def _sellevel(
--- a/aqua/core/backend/backend.py
+++ b/aqua/core/backend/backend.py
@@ -162,7 +162,9 @@ class Backend(ABC):
     def _seldate(self, data: xr.Dataset, startdate: str = None, enddate: str = None):
         """Store date bounds for lazy application."""
         t_start = pd.Timestamp(startdate) if startdate else None
-        t_end = pd.Timestamp(enddate) if enddate else None
+        # This guarantees that the end of year, day, hour etc is correctly selected
+        t_end = pd.Period(enddate).end_time if enddate else None
+        # t_end = pd.Timestamp(enddate) if enddate else None
         return data.sel(time=slice(t_start, t_end))
 
     def _sellevel(
--- a/aqua/core/backend/backend.py
+++ b/aqua/core/backend/backend.py
@@ -1,7 +1,7 @@
 from abc import ABC, abstractmethod
 
 import dask.array as da
-import numpy as np
+import pandas as pd
 import xarray as xr
 from smmregrid import GridInspector
 
@@ -161,7 +161,9 @@ class Backend(ABC):
 
     def _seldate(self, data: xr.Dataset, startdate: str = None, enddate: str = None):
         """Store date bounds for lazy application."""
-        return data.sel(time=slice(np.datetime64(startdate, "us"), np.datetime64(enddate, "us")))
+        t_start = pd.Timestamp(startdate) if startdate else None
+        t_end = pd.Timestamp(enddate) if enddate else None
+        return data.sel(time=slice(t_start, t_end))
 
     def _sellevel(
         self,
--- a/aqua/core/backend/backend.py
+++ b/aqua/core/backend/backend.py
@@ -11,6 +11,8 @@ from aqua.core.fixer import Fixer
 from aqua.core.logger import log_configure
 from aqua.core.util import fix_calendar, set_attrs, to_list
 
+_DEFAULT = object()
+
 
 class Backend(ABC):
     """
@@ -102,7 +104,7 @@ class Backend(ABC):
         if var:
             data = self._selvar(data=data, var=var)
         if startdate or enddate:
-            data = self._seldate(data=data, startdate=startdate, enddate=enddate)
+            data = self.seldate(data=data, startdate=startdate, enddate=enddate)
         if level:
             data = self._sellevel(data=data, level=level, level_coord=level_coord)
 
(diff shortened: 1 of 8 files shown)


aqua/core/backend/backend.py
```diff
@@ -160,7 +161,7 @@ def _select_minimum_sample(self, data, startdate: str = None):
 
     def _seldate(self, data: xr.Dataset, startdate: str = None, enddate: str = None):
         """Store date bounds for lazy application."""
-        return data.sel(time=slice(startdate, enddate))
+        return data.sel(time=slice(np.datetime64(startdate, "us"), np.datetime64(enddate, "us")))
```
mnurisso: precision of the time is something for which we ensure us as default precision, but it can be differently set in the catalog, so this feature should take into account different units set in the catalog metadata `time_coder`

oloapinivad: I agree, it cannot be hardcoded here. However I understand it might not be trivial to import it from another piece of the code.

oloapinivad: Minimum requirements is to import `DEFAULT_TIME_UNIT` and use pandas instead of numpy, but this is more complex that it seems

jhardenberg: Well ... it could be as simple as getting the precision from that of the data. Look into data, see which precision it is using and select in time using the same one. The original issue was that slice would use nanoseconds by default. In any case, this is probably an overkill: 
This piece of code should work no matter what. Using micorseconds here for the selection does no harm, because it will be used only for the selection...the result will still be in whatever precision data uses, no?

jhardenberg:
Apart from the discussion above, my other problem is another one: The failing test (which is not the one which I added here, that one works with the fix)  is because if I do `.sel("1985-01-01", "1985-12-31")` that will select everything till 1985-12-31T2359. If I use `np.datetime64` it will select till 1985-12-31T0000 (and the test is using data on 1985-12-31T2345 which get exclued). So the behaviour changed.
The question is what do we want? 
Of course one could decide that we have always to be precise in enddate and that if you wish your last day to be included you should make sure to specify as enddate "1985-12-31T2359" instead of "1985-12-31".  In this case the solution would be to fix the test, not the code.

On the other hand just specifying the last day could be useful (and is what has been assumed so far). In this case we would need to do something more complex where, if the enddate is a string not specifying hours, we should add one day - 1 second or similar to make sure to get the last day ... a bit ad hoc, not very elegant.

aqua/core/backend/backend.py
jhardenberg: I see that you guys also do not like the fact that I use numpy.datetime64  and would prefer to use pandas. I agree in principle but I have to check if pandas allows to specify the time units. I will test it

jhardenberg:
By the way, once we solve this here, there is a similar `.sel(time= ...)` issue also in diagnostics to be done afterwards

mnurisso:
> By the way, once we solve this here, there is a similar `.sel(time= ...)` issue also in diagnostics to be done afterwards

In the main issue is proposed the solution to this specific issue

aqua/core/backend/backend.py
jhardenberg: ok, the latest change simply uses pandas and apparently it does need at all that we specify time resolution, it will deal with any date also beyond 2262. Other advantage is that pandas is more flexible in understanding dates

jhardenberg:
My other problem instead remains (and for that reason tests are still failing). suggestions on that?

jhardenberg:
ok, I revised this in a way which addresses all worries discussed above:

1) we now use pandas
2) no need to specify time resolution (pandas works always)
3) we recover the exact behaviour of `.slice()` with strings so that if we specify enddate="1985-12-31" then also dates on "1985-12-31T23" will be included (pandas is wonderful)

All tests should pass now.

@oloapinivad @mnurisso please let me know what you think of it now.

jhardenberg:
Next issue: in AQUA-diagnostics also `diagnostic.py` (and a few diagnostics) contain `data = data.sel(time=slice(start, end))` or similar. What should we do?

Maybe it would make sense to add to the reader a `.seltime()` method to be used instead of `.sel(time=`... ? This will allow also to be smarter with selection of the time variable once the datamodel will be able to identify it. What do you think ?

In theory it should then become possible to do `data.aqua.seltime(startdate, enddate)` or `reader.seltime(data, startdate, enddate)`...

jhardenberg:
Ah, ok, I finally read #3072, I see you agree with my idea of a seltime/seldate (indeed the second name is better) method

mnurisso:
> Ah, ok, I finally read #3072, I see you agree with my idea of a seltime/seldate (indeed the second name is better) method

Yes, the only issue to be checked is if the `seldate` method (and the hidden `_seldate` from the backend) they can have access to the metadata mentioned defining the units precision. If this is not possible we agree `us` is ok, but please use the units defined in the `definitions.py` file, which is indeed `us` but defined with a global variable.

Also it would be ideal to define this new `seldate` method as accessor so that in the diagnostics we can do something like from `data.sel()` to `data.aqua.seldate()`.

Finally let us know if this modification is something you want to finish or we want to assign it to someone else, depending on your free time and need for this PR in the main.

jhardenberg:
@mnurisso actually the title of the issue has become irrelevant, using pandas we need not to worry about precision, it does it automatically

mnurisso:
> @mnurisso actually the title of the issue has become irrelevant, using pandas we need not to worry about precision, it does it automatically

that's very nice to know. However the relevance of a seldate method is still there right? So that we're sure diagnostics can use this centralized method

jhardenberg:
writing it ;)

aqua/core/reader/reader.py
```diff
@@ -717,6 +717,14 @@ def histogram(self, data, **kwargs):
         """
         return histogram(data, **kwargs)
 
+    # Other convenience wrappers for reader methods
+
+    def seldate(self, data, *args, **kwargs):
```
mnurisso, line 722: This are the docstring exposed in the API and which are most likely seen by a user. I'd populate them quite in detail, otherwise this method will remain totally obscure

mnurisso (approved):
See my comment on docstring, everything else seems fine

aqua/core/reader/reader.py
jhardenberg: You are totally right, but I notice that also all other wrappers (fldmean, fldsum etc ) do not have any docstrings and are badly documented in the docs. Maybe we should open an issue and make a separate PR to fix all that?

jhardenberg: Ah, ok, I see the logic: `timstat` and `fldstat` have extensive docstrings and the other ones need not to be repeated. Ok, I will fix the docstring

jhardenberg: ok @mnurisso I fixed the docstring, let me know if ok

mnurisso (approved):
It can be merged now in my opinion.

oloapinivad (approved):

PR #3079, opened 2026-09-09
New fix for ERA5

oloapinivad:
## PR description:

Following a change in the file shortnames in ERA5 dataset, we need to create a new fixer for this. 
Simply replacing upper case with lower case shortnames. 

In addition, clear-sky variables are finally recognized and their units correcly converted by the `Fixer`.


mnurisso (approved):

mnurisso:
Merging this should trigger a new release otherwise it will be impossible for python users to use the new updated era5 data

PR #3083, opened 2026-09-10
fix log_history message for flipped string

mnurisso:
## PR description:

Mini fix, without the message appears like:

('Flipped coordinate lat from ', 'decreasing to increasing by datamodel');

Notice parenthesis and extra ',' in the message


oloapinivad:
Ups, I found the same glitch and fixed it - bad habit - in this one: https://github.com/DestinE-Climate-DT/AQUA/pull/3079

oloapinivad (approved):

PR #3085, opened 2026-09-12
Seldate work with datetime

jhardenberg:
## PR description:

This is a small followup to #3072 to cover the marginal case where a datetime (or a pandas timestamp etc) is passed instead of a string as `enddate`. Normally if a string like "1985-12-31" is passed, the end of day is included (it becomes 1985-12-31T2359". If a datetime-like object is passed it is used as it is.

Related to AQUA-diagnostics  https://github.com/DestinE-Climate-DT/AQUA-diagnostics/pull/403

-----
 - [x] Changelog is updated.
 - [x] Test included

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ ClimateDT workflow modifications:
 
 Complete list:
 - Update Data Portfolio to v2.2.0 (#3084)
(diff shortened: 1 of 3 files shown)


tovaz92 (approved):
Approving, looks good

PR #3099, opened 2026-09-18
Add S3 bucket name validation relaxation for compatibility with LUMI-O

### LINKED ISSUE #3095: [Backend] Allow Lumi-O s3 bucket access
We have now LRA in zarr format on Lumi-O buckets.
The bucket name contains the project number with semicolumns, this is not a valid bucket name for botocore, which raise an error, for this reason a workaround is needed. 

After the merge of the new intake xarray drive #2958 we will add to the `intake_drivers/xarray/xzarr.py` file:

```python
def _relax_s3_bucket_name_validation():
    """Drop botocore's AWS-only bucket-name validator from its builtin handlers.

    Some S3-compatible object stores used for AQUA zarr sources (e.g. LUMI-O) use bucket
    names containing a colon, which are invalid on real AWS: botocore's built-in
    ``validate_bucket_name`` handler rejects the request before it is even sent. Removing
    the entry from ``BUILTIN_HANDLERS`` (read by every new (aio)botocore ``Session``, as
    created internally by s3fs on every filesystem instantiation) disables that check
    process-wide; the actual S3-compatible endpoint still validates the bucket itself.
    """
    try:
        from botocore import handlers as botocore_handlers
    except ImportError:
        return
    botocore_handlers.BUILTIN_HANDLERS = [
        entry for entry in botocore_handlers.BUILTIN_HANDLERS if entry[1] is not botocore_handlers.validate_bucket_name
    ]


_relax_s3_bucket_name_validation()
```

mnurisso:
## PR description:

The PR enable the lumi-o bucket access together with the possibility to have in the same source a S3 path and a local path depending on the machine. This allow to have S3 access on any machine while keeping the local access for the machine in which there is a physical copy of the data.

The PR is still missing a test because the S3 bucket access is introduced in this catalog PR: https://github.com/DestinE-Climate-DT/Climate-DT-catalog/pull/287

Close #3095

 - [ ] Changelog is updated.


aqua/core/intake_drivers/xarray/xzarr.py
```diff
@@ -37,6 +65,11 @@ def __init__(self, urlpath, storage_options=None, metadata=None, xarray_kwargs=N
         if "chunks" not in xarray_kwargs:
             kwargs.setdefault("chunks", {})
 
+        # zarr rejects storage_options on a plain local-path store: it is only honoured
+        # when the store is opened as an fsspec URI, so drop it for local urlpaths
+        if storage_options and not _is_fsspec_urlpath(urlpath):
```
jhardenberg, line 70: Looks good, but could you add a test for these lines to increase coverage?

jhardenberg (changes requested):

mnurisso:
@jhardenberg I needed to add the s3fs package in order to access the bucket, do we need to recreate the conda lock right?

mcadau (approved):

PR #3100, opened 2026-09-18
Adapt z3fdb to API change

### LINKED ISSUE #3098: z3fdb troubles
### machine

Lumi

### catalog

ci

### version

main

### What happened?

There are a couple of issues with using the z3fdb engine.

1) The backend filters unknown kwargs and considers `config_fdb` unknown. As a result, this code 

````
from aqua import Reader
reader = Reader(catalog='ci', model='IFS', exp='test-fdb', source="fdb", engine="z3fdb", config_fdb="/users/jvonhar/config.yaml")
data = reader.retrieve()
````

leads to 

````
2026-09-18 14:27:06 :: BackendIntakeFDB :: WARNING  -> kwarg config_fdb is not in the intake parameters of the source, removing it
````

2) In the latest z3fdb v 5.23.2.27 ECMWF has changed the API of `SimpleStoreBuilder.add_part()` which now accepts a dictionary and not a string for the request (!!)
So z3fdb breaks in newer environments

See https://sites.ecmwf.int/docs/fdb/5.23.5/z3fdb/api.html#z3fdb-simplestorebuilder

### PR DESCRIPTION
**Pull request by jhardenberg**
## PR description:

In a recent release of z3fdb (5.23.2.27) the API has changed and our driver requires some fixes. Additionally we noticed that the `config_fdb` argument is never passed to the driver since it is removed by the kwargs filtering routine.

Notice that this PR modifies the environment (and conda.lock)

Close #3098 

----

 - [x] Changelog is updated.

### CODE REVIEW

**Comment by jhardenberg** [comment]
for now the kwarg filtering is commented to fix the other part, it will be addressed next

**Comment by jhardenberg** [comment]
Ok, actually `BackendIntakeFdb` reintroduces the misisng kwarg after it has removed (this is a bit convoluted ... ) so the first point may not be an issue. The second one is being solve dby this PR

**Comment by jhardenberg** [comment]
Apart from the failing test being discussed in #3102 and which has nothing to do with this one, this seems ready to merge to me

**Comment by oloapinivad** [comment]
@jhardenberg this seems to work also without eccodes 2.48, is this something expected?

**Comment by jhardenberg** [comment]
> @jhardenberg this seems to work also without eccodes 2.48, is this something expected?

quite unexpected ... anyway, that's good

**Comment by jhardenberg** [comment]
I am doing some more tests

**Comment by mnurisso** [comment]
Ok, I was going to review it, let me know when it's ready

File: aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py [hunk 1]
```diff
@@ -43,8 +48,11 @@ def rebuild_fdb_zarr_store(config, mars, serialized_axes, extractor_type_str):
 
     builder = SimpleStoreBuilder(fdb_config_file)
     axes = []
-    for keys, chunking_str in serialized_axes:
-        chunking = getattr(Chunking, chunking_str)
+    for keys, chunking_spec in serialized_axes:
+        if isinstance(chunking_spec, (list, tuple)) and chunking_spec[0] == "FixedSizeChunk":
```
mnurisso, line 52: is there a possibility that we have a list/tuple and the `chunking_spec[0]` is not `FixedSizeChunk`? Is the else working in this case or do we need a raise?

aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py
```diff
@@ -217,27 +225,54 @@ def _build_mars_requests(request, freq, levels, years, start_date=None, end_date
...
     if freq in ("h", "D"):
-        time_axes = [AxisDefinition(["date", "time"], Chunking.SINGLE_VALUE)]
+        time_keys = ["date", "time"]
     elif freq == "MS":
-        time_axes = [AxisDefinition(["year", "month"], Chunking.SINGLE_VALUE)]
+        time_keys = ["year", "month"]
     else:
         raise ValueError(f"Unknown freq {freq!r}")
 
+    chunking = Chunking.SINGLE_VALUE
```
mnurisso: An extra line of comment on what is done would be helpful for maintenance

mnurisso (commented):
Seems good, waiting the ping for testing data access on lumi

aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py
jhardenberg: Indeed the check seems redundant because the two things happen always together. Since this structure is built by our code itself (see lines 538 - 550), there is no risk that this exotic situation occurs, so I would not worry with testing for it or with a raise (those lines would also be impossible to cover, since there is no test which we could write to trigger them). I would leave as is

aqua/core/intake_drivers/fdb/openers/z3fdb_opener.py
jhardenberg: done

jhardenberg:
wow, it all seems to work with the envirionment, ok. @mnurisso I will just wait for your review to finish then, for me this one is ready

jhardenberg:
We need to try and relock this one though ....

jhardenberg:
I tried doing mamba create of this environment on mafalda and it fails .....

mnurisso:
I'd test this PR on lumi since I need anyway to access the ICON data to be checked. Is the lumi-install to be updated here before merging it?

oloapinivad:
My understanding from https://github.com/DestinE-Climate-DT/AQUA/pull/3116 is that eccodes 2.48 is breaking our environment. no matter what. so we cannot merge this PR until we get what has changed. 

My investigation from the previous bug was that eccodeslib 2.48 is shipping a different dependency on PROJ which is not compatible with our current environment.

jhardenberg:
For testing: this sequence allows to reproduce the CI failure
`python -c "import z3fdb; import cartopy.crs"`

jhardenberg:
So it turns out that z3fdb 5.2.3.2.29 is problematic but that 5.3.2.27 instead works .... a pin seems to solve the issue.

jhardenberg:
This is now working using  eccodes 2.48.0 + eccodeslib 2.48.2.27 and fdb5lib/z3fdb 5.23.2.27

jhardenberg:
@mnurisso the env issues are solved (checked also with Paolo this morning) so if you can look if the rest of the PR is ok and give a final ok we can merge!

environment.yml
```diff
@@ -12,7 +12,7 @@ dependencies:
   # binaries
   - python>=3.11,<3.15
   - cdo>=2.5.0,<2.6.1 #for smmregrid compatibility
-  - eccodes>=2.41.0,<2.48
+  - eccodes>=2.41.0,<2.49
```
oloapinivad: missing the same modification for `environment-pypi.yml`

mnurisso (approved):
Code is ok (there is already a comment by Paolo on a missing modification). I'm testing on lumi for a final ok

mnurisso:
I've managed to create the env on lumi (the lumi env file has to be updated).
I can access data with this code:

```python
reader = Reader(
    catalog='climatedt-gen2',
    model='ICON-5km',
    exp='baseline-hist',
    source='monthly-hpz7-sfc',
    loglevel="INFO",
    areas=False,
    engine='z3fdb'
)
data = reader.retrieve()
```

however this code instead fails:

```python
reader = Reader(
    catalog='climatedt-gen2',
    model='ICON-5km',
    exp='baseline-abrupt4xco2',
    source='monthly-hpz7-sfc',
    loglevel="INFO",
    areas=False,
    engine='z3fdb'
)
data = reader.retrieve()
```

with this error:

```python
Exception: Failed system call: Read error on BufferedHandle[MultiHandle]: failed to read 1 byte (/src/eckit/src/eckit/io/PeekHandle.cc:54 peek) (Success)
Exception: UserError: Cannot create view, no data found for request class=d1,dataset=climate-dt,activity=baseline,experiment=abrupt4xco2,generation=2,model=ICON,realization=1,resolution=standard,expver=1,type=fc,stream=clte,param=235087/235088/235134/235136/235137/235151/235165/235166/228004/235168/228005/235079/235078/235020/235021/235031/235033/235034/235035/235036/235037/235038/235039/235040/235041/235042/235043/235049/235050/235051/235052/235053/235055,levtype=sfc,year=1995/1996/1997/1998/1999/2000/2001/2002/2003/2004/2005/2006/2007/2008/2009/2010,month=1/2/3/4/5/6/7/8/9/10/11/12 to establish field size. Underlying error: Failed system call: Read error on BufferedHandle[MultiHandle]: failed to read 1 byte (/src/eckit/src/eckit/io/PeekHandle.cc:54 peek) (Success) 
```

While with the 'gsv' engine it works.

I'd like to understand the issue before merging this.

environment.yml
jhardenberg: done, thx

jhardenberg:
> While with the 'gsv' engine it works.
> 
> I'd like to understand the issue before merging this.

Did it work with main and the old env ?

jhardenberg:
I updated env for lumi_install, even if this will soon be made obsolete by #3130

jhardenberg:
@mnurisso I notice that in the second case  (abrupt4xco2) z3fdb it is trying to read from stream clte (not clmn) using years and months to specify the date. Notice that instead in the case of baseline-hist `stream: clmn ` is used and works. z3fdb uses the frequency of data to decide if to use years/month or date/time to define time ... maybe that's the difference. In any case is there a difference in the catalog?

jhardenberg:
in any case this is a 'bug' (or feature) which we had already before, not related to this PR

jhardenberg:
the monthly data of abrupt4xco2 use `timestyle: clte` while those of baseline-hist use `timestyle: clmn`. Notice that the z3fdb driver ignores timestyle and uses simply the frequency to decide which timestyle to use. We can correct this in a separate PR but .... how come that abrupt4xco2 data are saved with `stream: clte` and not `stream: clmn` ????

PR #3102, opened 2026-09-20
[CI/CD] Test the CI/CD actions for issues

jhardenberg:
## PR description:

This is for now only a test because I got some strange CI/CD issues in #3100 and I want to check if they occur also in a clean branch


jhardenberg:
So the no-fdb action keeps failing with

````
2026-09-20 08:55:14 :: FileLock :: DEBUG    -> Acquired: /home/runner/.aqua/config-aqua.yaml.lock
2026-09-20 08:55:16 :: FileLock :: DEBUG    -> Released: /home/runner/.aqua/config-aqua.yaml.lock
free(): invalid pointer
/home/runner/work/_temp/c1d1de65-b1ba-46f9-86ba-ed904555d53e.sh: line 2:  4257 Aborted                 (core dumped) aqua -vv install github
Error: Process completed with exit code 134.
````

**Comment by jhardenberg** [comment]
This is quite mysterious.... I asked gemini and here is what it suggests. I will try to implement the suggestion later tonight/tomorrow morning

----

The failure in the GitHub Actions workflow "AQUA tests (without FDB)" (
aqua-no-fdb.yml) is caused by an unjoined daemon heartbeat thread in 
SafeFileLock that is still running while the Python interpreter is finalizing/shutting down, which triggers a heap memory corruption / double free or corruption (!prev) abort in glibc.
Here is the detailed breakdown:

1. The Core Bug: SafeFileLock hangs and leaves an active daemon thread

Look closely at the timestamps from the failure log:

```
2026-09-19 23:07:55 :: FileLock :: DEBUG    -> Acquired: /home/runner/.aqua/config-aqua.yaml.lock
2026-09-19 23:07:57 :: FileLock :: DEBUG    -> Released: /home/runner/.aqua/config-aqua.yaml.lock
double free or corruption (!prev)
Aborted (core dumped) aqua -vv install github

```

Notice that it took exactly 2.0 seconds between Acquired and Released.
In 
aqua/core/console/install.py:

```python

def _set_machine(self, args):
   ...
   with SafeFileLock(self.configfile + ".lock", loglevel=self.loglevel):
       cfg = load_yaml(self.configfile)
       cfg["machine"] = machine
       dump_yaml(self.configfile, cfg)

```

This is the very last operation executed by aqua install. As soon as this block exits, aqua finishes its execution and Python initiates interpreter shutdown (Py_FinalizeEx).
Now look at 
aqua/core/lock/safelock.py:


```python
def _heartbeat(self):
   """Refresh mtime periodically to indicate lock holder is alive."""
   while not self._stop_event.is_set():
       if os.path.exists(self.lock_path):
           try:
               now = time.time()
               os.utime(self.lock_path, (now, now))
           except Exception:
               pass
       time.sleep(self.heartbeat_interval)  # <--- DEFAULT IS 10 SECONDS!

def release(self):
   """Stop heartbeat and release the lock."""
   self._stop_event.set()
   if self._heartbeat_thread and self._heartbeat_thread.is_alive():
       self._heartbeat_thread.join(timeout=2)  # <--- 2-SECOND TIMEOUT!
   ...

```

When the lock is acquired, _heartbeat starts and immediately enters time.sleep(10).
The config file update completes in milliseconds, and release() is called.
release() calls self._stop_event.set(). However, time.sleep() does NOT wake up on threading.Event.set()—it will keep sleeping for the remaining ~9.9 seconds.
self._heartbeat_thread.join(timeout=2) blocks for the full 2 seconds, times out, and returns while the thread is still running.
The lock is released, and the main thread immediately exits.

3. Why double free or corruption (!prev) occurs
When the main thread finishes, Python invokes Py_FinalizeEx() to tear down the interpreter:
Daemon threads in Python are not joined by the runtime.
The interpreter unloads modules, deallocates thread-local states, destroys the GIL/locks, and frees internal memory pools.
While this teardown is underway, the orphaned daemon thread wakes up or its thread-local storage / stack is cleaned up concurrently with shared C-extensions (like C-libraries loaded by netCDF4, icechunk, eccodes, cdo, libgcc/libstdcxx).
glibc's free() detects that a heap chunk header has been corrupted or freed concurrently, triggering double free or corruption (!prev) and SIGABRT (exit code 134).

4. Why did this happen only in aqua-no-fdb?
Host runner vs Docker container:
In 
aqua.yml, the workflow uses a container: container-image: ghcr.io/destine-climate-dt/ubuntu26.04-fdb5.19.0-eccodes2.47.0-aqua:aqua-base-container.
In 
aqua-no-fdb.yml, container-image: '' is set, meaning it runs directly on the GitHub Actions host runner (ubuntu-latest) with host glibc.
Python 3.14 & Recent Conda Relock:
Both workflows are running Python 3.14 (python-versions: '["3.14"]').
The relock on relock-conda-12 (commit 4a2038e1) updated C/C++ packages (libgcc, libstdcxx, libglib, icechunk, numcodecs, matplotlib).
On the ubuntu-latest host with Python 3.14, having active background threads during interpreter shutdown exposes glibc heap corruption upon process exit.

The Fix
In 
aqua/core/lock/safelock.py, _heartbeat() should wait on the event itself (self._stop_event.wait(self.heartbeat_interval)) instead of time.sleep():


```python
   def _heartbeat(self):
       """Refresh mtime periodically to indicate lock holder is alive."""
       while not self._stop_event.is_set():
           if os.path.exists(self.lock_path):
               try:
                   now = time.time()
                   os.utime(self.lock_path, (now, now))
               except Exception:
                   pass
           # Wait for interval OR wake immediately when stop_event is set
           if self._stop_event.wait(self.heartbeat_interval):
               break

```

This ensures that the moment self._stop_event.set() is called in release(), the heartbeat thread terminates in milliseconds, join() succeeds cleanly, and no background daemon thread remains running when Python exits.

oloapinivad:
The fact that this is not happening in FDB tests, where we use our own container, suggest an external change. ubuntu container has not changed afaik. However, the environment is fixed by the conda-lock. are there any other degree of freedom?

jhardenberg:
Indeed, I agree. 

So the solution suggested by AI (removing hartbeat) does not work. I will use this PR as a playground and try another approach then, even if we need then to discuss if we like something like that

oloapinivad:
Check the issue, I am able to replicate it locally. it is the environment, not the code. I am suspecting condalock honestly

jhardenberg:
My AI says the following:

The Python code is completely valid Python.

There is no Python syntax error, no uncaught exception, and no logic bug in AQUA. If this were a pure Python issue, Python would either complete normally or print a standard Python traceback.

Here is why this error occurs despite the Python code being valid:

1. What double free or corruption actually means
This message is not coming from Python. It is printed directly by glibc (the GNU C standard library, libc.so.6) inside its heap memory allocator (malloc / free).

Glibc aborts the process (SIGABRT, exit code 134) when low-level C code:

Calls free(pointer) on a memory address that has already been freed, or
Writes past the boundary of a buffer, overwriting the internal memory management metadata (chunk->prev_size, hence !prev) of an adjacent chunk on the heap.
Pure Python code cannot cause this. It can only be caused by compiled C, C++, Fortran, or Rust binary extensions.

2. Why does a Python script contain compiled C/C++ code?
Python itself is a C program (CPython). Furthermore, scientific Python libraries are almost entirely C/C++/Fortran/Rust wrappers:

numpy / scipy: C, Fortran, OpenMP (libgomp)
netCDF4 / HDF5: C
eccodes / cdo: C, C++
cartopy / shapely / geos / proj: C++
matplotlib / contourpy: C++
icechunk: Rust compiled with PyO3
When Python exits (exit()), it executes Py_FinalizeEx() to tear down the runtime:

It runs atexit callbacks.
It deletes modules in sys.modules, triggering Python __del__ object destructors.
The dynamic linker (ld.so) executes C++ global destructors and .so exit handlers in reverse order.
If any single C/C++ extension has a memory corruption bug, an invalid destructor, or a conflict between allocators, glibc halts the entire process with double free or corruption (!prev).

3. Why did this start happening now, and why only in aqua-no-fdb?
It worked before, so what changed? Three things converged:

A. Python 3.14 internals
Python 3.14 is very new and introduces major architectural changes to CPython's memory management (deferred reference counting, new garbage collection mechanisms, mimalloc integration). Several C/Rust extensions are still adapting their finalizers to 3.14.

B. Mixing PyPI wheels and Conda-Forge binaries
In aqua-test-reusable.yml, the environment installs Conda packages and then pip-installs wheels from PyPI:

bash
sed -n 's/^# pip //p' conda-linux-64.nofdb.lock | python -m pip install -r /dev/stdin
Wheels like icechunk-2.2.2-cp312-abi3-...whl are pre-compiled with generic Rust/PyO3 against the Python 3.12 Stable ABI (abi3).
Conda packages (libstdc++, cdo, libglib) are compiled by Conda-Forge with GCC. When running under Python 3.14, ABI3 Rust/C++ extensions interacting with Python 3.14's new runtime cleanup can double-free internal objects during Py_FinalizeEx.
C. Host Runner vs Docker Container
In aqua.yml, tests run inside a pinned Docker container (ubuntu26.04-...-aqua-base-container).
In aqua-no-fdb.yml, container-image: '' is set, meaning it runs directly on the GitHub Actions host runner (ubuntu-latest). On the host runner, Conda packages run against Ubuntu's host glibc (glibc 2.39), exposing memory cleanup conflicts that did not trigger inside the container.
D. The relock-conda-12 updates
The conda relock on Sep 18 updated:

libglib from 2.88.3 to 2.90.0
icechunk to 2.2.2
libgcc / libstdc++ to build _5
matplotlib-base to 3.11.2
One of these native libraries introduced an exit destructor bug under Python 3.14 + host glibc.

4. Why does Step 2 matter?
aqua install has no business loading matplotlib, cartopy, cdo, eccodes, netcdf4, or icechunk into memory in the first place.

The reason they were loaded was purely accidental:



aqua/core/console/parser.py
 did from aqua import __version__ instead of from aqua.core.version import __version__.


aqua/core/console/main.py
 imported analysis_execute, builder_execute, drop_execute, and catgen_execute at module top-level.
Because of this, merely typing aqua -vv install github loaded over 50 C/C++ shared libraries into memory, only to copy a couple of YAML files and exit. When it exited, one of those 50 C libraries crashed in its destructor.

By making those CLI imports lazy:

aqua install (and aqua list, aqua add, aqua set) only loads Python's standard YAML/OS modules and runs cleanly in 0.05 seconds.
The CI pipeline will proceed past the setup step into pytest.
When pytest runs the tests that actually use those scientific libraries, if an individual library is broken under Python 3.14, pytest will isolate and report exactly which library/test triggers the issue.


----

the "step 2" suggestion is a suggestion (from a previous interaction) wher eit would like to change radically where we import libraries in the code

oloapinivad:
This is supersed by #3107: it can be closed

jhardenberg:
ok, closing!

